PR #18, opened 2015-01-26
add match for souce object

lexich:
simple improve of previous PR


MoOx:
In why this is improving ?
Any tests to add?

lexich:
I'm not sure that source object can be empty otherwise, It's a part of internal API.

PR #23, opened 2015-02-11
Create `before` property on `first` element, only if defined

### LINKED ISSUE #22: TypeError: Cannot read property 'before' of undefined
With this CSS:

```css
@import url(http://fonts.googleapis.com/css?family=Noto+Sans:400,700,400italic,700italic);
```

The error occurs on `first.before = "\n\n" + first.before`  from `index.js:112`

Same error using postcss-import v4 with postcss v3.

Related: iamvdo/pleeease#40

iamvdo:
It fixes #22, but as there is no comment, I don't know why you try to clone `before` with `\n\n`

### DIFF
--- a/index.js
+++ b/index.js
@@ -109,7 +109,9 @@ function addIgnoredAtRulesOnTop(styles, ignoredAtRules) {
       styles.nodes.unshift(ignoredAtRule)
     }
 
-    first.before = "\n\n" + first.before
+    if (first) {
+      first.before = "\n\n" + first.before
+    }
   }
 }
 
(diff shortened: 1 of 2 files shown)


MoOx:
Just to add a separation, otherwise you can get (from what I remember) 

Input:

```
@import "a.css";
@import "b.css";
```

```
a{}b{}
```

But 


```
a{}

b{}
```

make more sense (to me)

Thanks for this.

MoOx:
Can't merge a PR without test added to cover the bug fixed.

iamvdo:
I know, but I'm not sure what to test ? (there wasn't a test for that before...)

I can add the sample code as a test, but is it sufficient and bring benefits ?

MoOx:
Benefits fo the sample: ensure the bug don't reappears again in the future :)
You can just modify an existing sample in current head, ensure test is broken, then add your fix, then ensure test pass. That's how I fix bugs :)

MoOx:
Released as 5.0.1

PR #24, opened 2015-02-12
Properly clone nodes in `ignoredAtRules`

### LINKED ISSUE #21: TypeError when using @import for local and remote stylesheets
If I have one or more `@import` pointing to an external URL (e.g. `@import url("//hello.myfonts.net/count/");`) and nothing else, postcss-import doesn't inline them. :thumbsup:

However, if the stylesheet *also* has an `@import` pointing to a local stylesheet for inlining, I get the following:

```bash
/Projects/postcss-import-test/node_modules/postcss/lib/node.js:284
                if (p != root && p.parent && p.parent == root) {
                                  ^
TypeError: Cannot read property 'parent' of undefined
    at /Projects/postcss-import-test/node_modules/postcss/lib/node.js:284:35
    at /Projects/postcss-import-test/node_modules/postcss/lib/container.js:116:26
    at Root.each (/Projects/postcss-import-test/node_modules/postcss/lib/container.js:94:22)
    at Root.eachInside (/Projects/postcss-import-test/node_modules/postcss/lib/container.js:115:21)
    at Declaration.style (/Projects/postcss-import-test/node_modules/postcss/lib/node.js:282:18)
    at Declaration.style (/Projects/postcss-import-test/node_modules/postcss/lib/node.js:362:35)
    at Declaration.stringify (/Projects/postcss-import-test/node_modules/postcss/lib/declaration.js:19:27)
    at Rule.stringifyContent (/Projects/postcss-import-test/node_modules/postcss/lib/container.js:34:27)
    at Rule.stringifyBlock (/Projects/postcss-import-test/node_modules/postcss/lib/container.js:49:18)
    at Rule.stringify (/Projects/postcss-import-test/node_modules/postcss/lib/rule.js:23:14)
```

Also, if the stylesheet to be inlined only contains an empty declaration block (e.g. `p {}`), I don't get the error but do see:

```css
@import url("http://hello.myfonts.net/count/");

undefinedp{}
```
Looks like something is creeping in with parsing, prepending that `undefined`, but haven't been able to get further than that.

Noticed this when bumping a project from `postcss@3.0.0` and `postcss-import@4.1.1` to `postcss@4.0.3` and `postcss-import@5.0.0`. Works as expected (ie. remote stylesheet `@import` ignored, local inlined) with the older versions.

Test case at [this gist](https://gist.github.com/ohskylab/5ea14e230068bcd8cfe4).

iamvdo:
(fix #21 when `parent` is undefined)

### DIFF
--- a/index.js
+++ b/index.js
@@ -109,7 +109,7 @@ function addIgnoredAtRulesOnTop(styles, ignoredAtRules) {
       styles.nodes.unshift(ignoredAtRule)
     }
 
-    if (first) {
+    if (first && first.before !== undefined) {
       first.before = "\n\n" + first.before
     }
   }
@@ -126,13 +126,17 @@ function readAtImport(atRule, options, cb, importedFiles, ignoredAtRules, media)
   // @todo extract what can be interesting from this one
   var parsedAtImport = parseImport(atRule.params, atRule.source)
 
+
+
   // adjust media according to current scope
   media = parsedAtImport.media ? (media ? media + " and " : "") + parsedAtImport.media : (media ? media : null)
 
   // just update protocol base uri (protocol://url) or protocol-relative (//url) if media needed
   if (parsedAtImport.uri.match(/^(?:[a-z]+:)?\/\//i)) {
     parsedAtImport.media = media
-    ignoredAtRules.push([atRule, parsedAtImport])
+    var atRuleCloned = atRule.clone()
+    atRuleCloned.parent = atRule.parent.clone()
+    ignoredAtRules.push([atRuleCloned, parsedAtImport])
     atRule.removeSelf()
     return
   }
--- a/test/index.js
+++ b/test/index.js
@@ -63,8 +63,19 @@ test("@import", function(t) {
       "should not fail with only one absolute import"
   )
 
+  base = "@import url('http://');\n@import 'test/fixtures/imports/foo.css';";
+  t.equal(
+    postcss()
+      .use(atImport())
+      .process(base)
+      .css.trim(),
+      "@import url('http://');\nfoo {}",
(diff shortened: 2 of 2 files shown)


index.js
```diff
@@ -126,13 +126,17 @@ function readAtImport(atRule, options, cb, importedFiles, ignoredAtRules, media)
...
   // adjust media according to current scope
   media = parsedAtImport.media ? (media ? media + " and " : "") + parsedAtImport.media : (media ? media : null)
 
   // just update protocol base uri (protocol://url) or protocol-relative (//url) if media needed
   if (parsedAtImport.uri.match(/^(?:[a-z]+:)?\/\//i)) {
     parsedAtImport.media = media
-    ignoredAtRules.push([atRule, parsedAtImport])
+    var atRuleCloned = atRule.clone()
+    atRuleCloned.parent = atRule.parent.clone()
+    ignoredAtRules.push([atRuleCloned, parsedAtImport])
```
MoOx: Why do we need this weird "clone but not really" ?

iamvdo: I found out that `atRule.removeSelf()` changes the rule previously stored in `ignoreAtRules`. So I chose to clone node before storing it. I also cloned `parent` property because the `clone()` method doesn't make it.
Do you think of another way to do it ?

MoOx: I don't really know (or remember) why we do need a parent ref, but anyway, this is fixing a bug, with a test, so merging ^^

iamvdo: Yep. Can you release a patch ?

MoOx: I am working a more stable release, but encountering too many weird behavior. This postcss API is starting to break my balls.

I will release a patch asap.

iamvdo: OK ;)

PR #26, opened 2015-02-18
Remove unnecessary if requirements

vanesyan:
It's unnecessary to check if `styles` existing and all it props separately to `styles.nodes[0].source.input.file`. If `style` will not exist so it already except the `if` statement

### DIFF
--- a/index.js
+++ b/index.js
@@ -41,7 +41,7 @@ function AtImport(options) {
 
   return function(styles) {
     // auto add from option if possible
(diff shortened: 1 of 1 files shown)


MoOx:
What if styles came from an empty stylesheet ? 💥
What if first node has been created manually & do no have any source ? 💥

vanesyan:
it'll skip this block, because there is no style, so there are no nodes[0], source, input, file.. That okay.

MoOx:
Please rebase on master, and run tests. Tell me what happen.

MoOx:
Also http://requirebin.com/?gist=bb5cdcf3ab1dca61465c

PR #27, opened 2015-02-19
Use common PostCSS plugin API

borodean:
Similar to this https://github.com/borodean/postcss-assets/issues/6
Reference: https://github.com/postcss/postcss-plugin-boilerplate

### DIFF
--- a/index.js
+++ b/index.js
@@ -23,6 +23,9 @@ var moduleDirectories = [
  * Expose the plugin.
  */
 module.exports = AtImport
+module.exports.postcss = function(styles) {
+  return module.exports()(styles)
+}
 
 /**
  * Inline `@import`ed files


MoOx:
😥
I'm totally against this horrible way to deal with plugin without arguments.
It's seems so inconsistent.

borodean:
@MoOx I dont agree with you. It is incosistent to have it not like the other plugins have.

So basically whem I’m adding other plugins to my script using this little syntax sugar, I cant do the same with `postcss-import`, which looks incosistent indeed:

```js
postcss([
    require('autoprefixer-core'),
    require('postcss-nested'),
    require('postcss-mixins'),
    require('postcss-import')()
])
```

MoOx:
Well it's just 3 lines.

PR #28, opened 2015-03-26
Not import same hash files

magsout:


index.js
```diff
@@ -188,10 +191,20 @@ function readImportedContent(atRule, parsedAtImport, options, resolvedFilename,
     return
   }
 
+  var fileContentHash = hash(fileContent)
+  // skip files already imported with the same content but different scope
+  if (!media && hashFiles.indexOf(fileContentHash) != -1) {
```
MoOx: Maybe we should extrapolate for media queries too.
See https://github.com/postcss/postcss-import/blob/master/index.js#L58-L60 & https://github.com/postcss/postcss-import/blob/master/index.js#L158
=> We should not only save the ref in a simple array but do like the importedFile array and store an object with mq ref to skip for each scope too.
If you don't feel comfortable with this modification, I can handle it. You already did 90% of the job ;)

magsout:
I turned my array in object, and makes the same tests as you. @MoOx

MoOx:
Released as 5.1.0

I've added notes in the README about that + more tests for scopes ;)

Thanks for this enhancement!

magsout:
:+1:  thanks for your help

PR #33, opened 2015-04-10
 Files which only contain same @import rules were skip

### LINKED ISSUE #31: Optimization by file content must skip "proxy" files
I've a main css file where I do:

```css
@import "module-a";
@import "module-b";
```

Both modules have an `index.css` with `@import "./src/styles/index.css";`

If I compile, it only imports the first module, not the second. For instance, if I switch from `a` to `b` like that:

```css
@import "module-b";
@import "module-a";
```

`module-b` will be imported but not `module-a`.

To fix it, I have to do that:

```css
@import "module-a/src/styles/index.css";
@import "module-b/src/styles/index.css";
```

I'm inspecting this bug but if you have already an idea why, I'll be glad to hear.

magsout:
Closed #31


MoOx:
I will add one more test for this and release that.

PR #36, opened 2015-04-16
Ensure glob works when using single quotes and/or url(path)

kswedberg:
Sorry, my previous PR only accounted for @import rules with double quotes. This commit ensures that globs work when the @import uses single quotes and/or the url(path) syntax. Added a test for it, too.


MoOx:
Very nice. Thanks.

MoOx:
Released as 5.2.1

PR #37, opened 2015-04-18
Make globbed imports work for module directories, too.

kswedberg:
Concatenated `moduleDirectories` to paths for globbed @import rules and modified the glob test to make sure it works. Wish I had noticed this the first time around. Anyway, this PR should be the last for the glob option. Thanks for merging these in so quickly. Really appreciate it!


MoOx:
Should I release this as a patch ?

kswedberg:
Yes, I think so. The previous commits didn't mention anything about glob patterns only working for local files, so I think the assumption would have been that they worked for module directories as well. Therefore, this is a fix, in a sense, for it "breaking" when trying to use it with module directories.

kswedberg:
Thank you!

MoOx:
Released as 5.2.2.

Thanks a lot for this feature.

PR #47, opened 2015-06-04
Add cache

andyjansson:
Added a `cacheDir` option. When set, it will enable the cache functionality and store the compiled output(s) along with a dependency graph in the specified directory. Modifying any of the files in the dependency graph invalidates the cache for that file and any dependents. Any unmodified dependencies will continue to be cached.


test/fixtures/cwd.css
```diff
@@ -1,2 +1,2 @@
-@import "test/fixtures/imports/foo.css";
-@import "test/fixtures/imports/foo-recursive.css";
+@import "imports/foo.css";
+@import "imports/foo-recursive.css";
```
MoOx: this test should not be changed since the point of this one is to check that without from option, you can use a file from your cwd (which is the root of this project when running test from the root)

index.js
```diff
@@ -83,7 +99,16 @@ function AtImport(options) {
  * @param {Object} styles
  * @param {Object} options
  */
-function parseStyles(styles, options, cb, importedFiles, ignoredAtRules, media, hashFiles) {
+function parseStyles(styles, options, cb, globallyImportedFiles, ignoredAtRules, media, hashFiles, cache) {
+  if (options.from && !media && cache && IsCached(options.from, options, cache)) {
```
MoOx: so you are not caching import that use a media query ? why ?

test/index.js
```diff
@@ -10,13 +10,16 @@ var postcss = require("postcss")
...
 var importsDir = path.join(fixturesDir, "imports")
+var cacheDir = path.join(__dirname, "cache")
 
 function read(name) {
   return fs.readFileSync("test/" + name + ".css", "utf8").trim()
 }
 
 function compareFixtures(t, name, msg, opts, postcssOpts) {
-  opts = opts || {path: importsDir}
+  opts = opts || {path: importsDir, cacheDir: cacheDir}
```
MoOx: why are you enabling the cache for all tests ?

index.js
andyjansson: Media queried imports will be cached by the requesting file. One shouldn't cache the file when requested with a media query because it will alter the output from how it looks when requested normally.

test/fixtures/cwd.css
andyjansson: Then I propose that the documentation should be altered to reflect that: 

>```css
>// `from` option is required so relative import can work from input dirname
>```

MoOx: it say "input", not "cwd" which is correct. If you make from a file `import "./stuff"`  and provide the from option, the file might be found in the same dir as the original input, and this cannot be done if you don't provide the from option.
What is not correctly clear ?
There is a test for that https://github.com/postcss/postcss-import/blob/debda623cd0d5354664b4412b87c29de1c16e6b6/test/index.js#L62

index.js
MoOx: hum... ok :)

andyjansson:
ping @MoOx

test/index.js
```diff
@@ -190,3 +192,22 @@ test("works with no styles at all", function(t) {
...
+  var expected = read("fixtures/" + name + ".expected")
+  t.equal(actual, expected, "put content in cache")
+  actual = postcss()
+    .use(atImport(opts))
+    .process(read("fixtures/" + name), postcssOpts)
+    .css.trim()
+  expected = read("fixtures/" + name + ".expected")
+  t.equal(actual, expected, "read content from cache")
+  t.end()
+})
```
MoOx: how is this testing that the cache has been used ?!

andyjansson: The first `process()` call puts the compiled output [into the cache](https://github.com/andyjansson/postcss-import/blob/master/index.js#L128-140) since `cacheDir` has been specified.

The second `process()` call rather than compiling the CSS again instead [reads it from the cache](https://github.com/andyjansson/postcss-import/blob/master/index.js#L103-111).

MoOx: The test does not prove anything right now :/. We should probably add something to track the cache usage.

andyjansson:
@MoOx The test has been updated to include a check of the cache's state.

test/index.js
```diff
@@ -190,3 +192,27 @@ test("works with no styles at all", function(t) {
...
+  var output = postcss()
+    .use(atImport(opts))
+    .process(css, postcssOpts)
+    .css.trim()
+
+  var imports = JSON.parse(fs.readFileSync(cacheDir + "/imports.json"))
+  var cacheFileName = imports[cssFileName].cache;
+  var cache = fs.readFileSync(cacheFileName, "utf8").trim()
+
+  t.equal(output, cache, "should put output in cache")
```
MoOx: we know here that the cache is written, how do we know the cache is being read after that ?

andyjansson: I'm sorry @MoOx, there's no way of knowing. It's a transparent cache. Unless you're suggesting that we should add a bunch of debug code to poll the internal state of the engine, which I would strongly advice against, we might as well close the PR because there's no way for me to satisfy this need. I know that the cache is hit by using breakpoints, I know that it returns a correct and consistent result, and I know that I get a 100ms performance increase when using it with a local project of mine, but there's no way for me to put that into a unit test.

MoOx: There is always a way. What about adding a simple map object with {file-requested: bool flag} to know if a requested file as been used from cache? Doesn't seems hard to do and this won't affect perfs.
If you don't know how to do it, I can probably handle that.

andyjansson: I don't quite understand what you mean, but you're welcome to take a stab at it.

MoOx:
I was about to merge that to handle the missing test but we made some changes in the master. Did you want to handle the rebase? Otherwise I will try to do it on tuesday.

jedmao:
@andyjansson you gonna' rebase this one?

andyjansson:
@jedmao I tried. I give up. I'll leave this to @MoOx

andyjansson:
@MoOx @jedmao I have a working version based on the latest source, however I haven't rebased the project. I'm thinking it might be easier to just create a new fork and add the code there. What do you think?

jedmao:
@andyjansson did you see https://github.com/postcss/postcss-import/pull/62?

andyjansson:
@jedmao you should probably PR to my repo

andyjansson:
Don't merge this yet. I need to verify that the cache works as intended.

andyjansson:
@jedmao Enabling caching for all tests and running the tests twice causes `should be able to consume npm package or local modules` to fail. It looks like it isn't ignoring an `@import` declaration for an already imported document. Having another eye on this issue would be swell.

.gitignore
```diff
@@ -1,2 +1,4 @@
 node_modules
+test/cache
 test/fixtures/*.actual.css
+test/cache
```
jedmao: We have a duplicated line here that needs to be removed.

jedmao:
Are you familiar w/ how to rebase? The merge commits in this PR are, I think, something we want to avoid.

jedmao:
See https://github.com/postcss/postcss-import/pull/63 for a clean rebase w/o merge commits. Feel free to submit a PR to [my cache branch](https://github.com/jedmao/postcss-import/tree/cache) with the failing test you're talking about. I'm not sure what the issue is there.

andyjansson:
@jedmao It's not ready to be merged, so there would be no point in squashing the commits before it is. Once the issue mentioned earlier has been resolved I'll do an interactive rebase.

jedmao:
I'd like to see a failing build w/ the issue you're speaking of.

andyjansson:
@jedmao Just change line 19 of `test/index.js` to the following and run `npm test` **twice**: 

```js
  opts = assign({path: importsDir, cacheDir: cacheDir}, opts || {})
```

jedmao:
@andyjansson I see what you mean now. Unfortunately, I don't know enough about how the caching works to really offer any help. My involvement in this PR was in response to https://github.com/postcss/postcss-import/pull/59#issuecomment-113865802, to get the ball rolling with my index.js refactor.

andyjansson:
@jedmao Well, I finally had time to fix the bug, so the good news is that everything is working. The bad news is that the fork is now behind the upstream repo, which is why appveyor is reporting an error. Also, 0cfa80687a160c463a549d95688ba14a4186f775 is a bit of a problem and might need to be reversed.

poeschko:
May I ask why this was closed? Does it not work correctly, would it need more work on the PR (rebasing it again), or do you think it would not provide any significant advantage over the improvements that have already been made? It seems to me that #152 only achieves parts of what a more general caching approach would (only caching file I/O, but not any transformations applied after that).

Using postcss-devtools, I found that postcss-import takes ~20 seconds in my project (accumulated over multiple CSS entry points), which is the vast majority of the overall build time. So any speedup would be great, and I'd be motivated to help if desired.

RyanZim:
@poeschko This PR was closed because of merge conflicts. Since then, the entire codebase was rewritten. No one ever attempted to replicate this feature.

I'd welcome any performance improvements.

There's two big performance problems in postcss-import:
- Poor Caching, only caches file I/O, not processing
- This line: https://github.com/postcss/postcss-import/blob/master/index.js#L177. That promiseEach serializes a lot of stuff that could run concurrently (in theory). It was added as a quick fix for a bug in the global deduping, but there's probably a more performant way to do it.

I'm not sure which problem is the biggest slowdown.

andyjansson:
@poeschko I closed the PR because I felt the feature was never going to get merged in, so I switched my focus to other things.

Acidic9:
This feature would be amazing. Especially in my case because I am importing src from a node module which adds a huge delay to my app.

AndyOGo:
This would be a great addition.

PR #52, opened 2015-06-14
allow custom extensions to be used when resolving paths

zoubin:
I'm using postcss to replace node-sass, and postcss-import seems to be greatly helpful. It will be more convenient if file extensions can be specified (i.e. `.scss`).


MoOx:
Not sure if we should merge this. It doesn't make sense to import Sass code in PostCSS which doesn't handle all the Sass syntax.
Do you know you can use the full path instead ? Like "path/to/file.scss" ? This should work.

zoubin:
@MoOx Oh, didn't know that. Thanks.
Actually, it is even better if there is a `options.resolve` to specify the resolve behaviour (rather than the default node-resolve.sync). Do you have any plans for that?

MoOx:
No because nobody ask for it. But it might be a more flexible solution indeed.

zoubin:
I am working on some stuff to pack css, which need to resolve dependencies. If I am going to use postcss-import, I have to make sure that my resolve-logic is the same with that in postcss-import. It really helps if the resolve logic in postcss-import can be configured.
So, would you consider supporting that in some future version, perhaps?

MoOx:
I am not against an option indeed.

zoubin:
Support custom resolve: https://github.com/postcss/postcss-import/pull/53

PR #53, opened 2015-06-16
custom resolve function option

zoubin:
Allow to specify the resolve behaviour by setting options.resolve


MoOx:
Cool thanks !
I will handle some issue and release that right after.

Issue #41, opened 2015-05-31
Async FS calls to improve perfomance


ai:
PostCSS 4.1 brings async support, so we can make `postcss-import` [much faster](https://github.com/postcss/postcss/issues/341)

andyjansson:
Added a [PR for caching](https://github.com/postcss/postcss-import/pull/47) which should also help with performance.

jedmao:
I have a commit [here](https://github.com/jedmao/postcss-import/commit/3a788c06a679f5cfab3a5eb92ffaa24a1db30cb1) you can cherry-pick to get started. Unfortunately, that code makes the `ignore` test fail and I can't yet figure out why. I think I've exhausted my willingness to work on this issue, but feel free to take that commit and do what you will with it.

Also, I noticed the use of `fs.existsSync` in this plugin's code. I want to point out the fact that [this will be deprecated](https://nodejs.org/api/fs.html#fs_fs_existssync_path) along with [the asynchronous version](https://nodejs.org/api/fs.html#fs_fs_exists_path_callback), `fs.exists`. There's also [another comment in the docs](https://nodejs.org/api/fs.html#fs_fs_exists_path_callback) that states:

> In particular, checking if a file exists before opening it is an anti-pattern that leaves you vulnerable to race conditions: another process may remove the file between the calls to fs.exists() and fs.open(). Just open the file and handle the error when it's not there. 

> fs.exists() will be deprecated.

Maybe the more graceful solution is to just try opening the files and throw any errors that come up.

MoOx:
6.1.1 has an `async` option to enable postcss async use, so we might use it to detect async mode and use async fs read. Should not be a big deal from here. Maybe we might use denodeify to transform fs.readFile to a promise.

MoOx:
I guess we can close, right @TrySound ?

PR #55, opened 2015-06-19
Introduce "plugins" option

### LINKED ISSUE #45: Add option to run a processor on imported AST
Convenient to run a linter on the fly. No extra processing.
Ideas:
- change onImport option to pass all files + latest filename/ast
- add an afterParse (or onParse) option that take filename + ast

jedmao:
Fixes #45

Willing to rename this option to something else, but I would appreciate your code review, @MoOx. I think I still need to write one more test that verifies warnings are being displayed for plugins applied on import.

### DIFF
--- a/index.js
+++ b/index.js
@@ -12,6 +12,11 @@ var helpers = require("postcss-message-helpers")
 var hash = require("string-hash")
 var glob = require("glob")
 
+var Promise = global.Promise || require("es6-promise").Promise
+var resolvedPromise = new Promise(function(resolvePromise) {
+  resolvePromise()
+})
+
 /**
  * Constants
  */
@@ -72,22 +77,33 @@ function AtImport(options) {
 
     var hashFiles = {}
 
-    parseStyles(
+    return parseStyles(
       result,
       styles,
       opts,
-      insertRules,
       importedFiles,
       ignoredAtRules,
       null,
-      hashFiles
-    )
-    addIgnoredAtRulesOnTop(styles, ignoredAtRules)
+      hashFiles,
+      createProcessor(result, options.plugins)
+    ).then(function() {
+      addIgnoredAtRulesOnTop(styles, ignoredAtRules)
 
-    if (typeof opts.onImport === "function") {
-      opts.onImport(Object.keys(importedFiles))
-    }
+      if (typeof opts.onImport === "function") {
+        opts.onImport(Object.keys(importedFiles))
+      }
+    })
   }
+}
+
+function createProcessor(result, plugins) {
+  if (plugins) {
+    if (!Array.isArray(plugins)) {
+      throw new Error("plugins option must be an array")
+    }
+    return postcss(plugins)
+  }
+  return postcss()
 }
 
 /**
@@ -100,11 +116,11 @@ function parseStyles(
   result,
   styles,
   options,
-  cb,
   importedFiles,
   ignoredAtRules,
   media,
-  hashFiles
+  hashFiles,
+  processor
 ) {
   var imports = []
   styles.eachAtRule("import", function checkAtRule(atRule) {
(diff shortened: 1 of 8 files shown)


index.js
```diff
@@ -71,22 +76,33 @@ function AtImport(options) {
...
+    })
+  }
+}
+
+function createProcessor(result, plugins) {
+  if (plugins) {
+    if (Array.isArray(plugins)) {
+      return postcss(plugins)
     }
+    result.warn("plugins must be an array")
```
MoOx: We should probably throw an error instead for this.

test/index.js
```diff
@@ -54,10 +59,16 @@ test("@import", function(t) {
...
 
   compareFixtures(t, "transform", "should support transform", {
-    path: importsDir,
     transform: require("css-whitespace"),
   })
 
+  compareFixtures(t, "plugins", "should support plugins", {
+    plugins: [
+      require("postcss-nested"),
+      require("postcss-bem-linter"),
```
MoOx: instead of adding dev deps (real plugins) we should code & use a stupid plugin with more stupid fixtures

MoOx:
I merged another pr so you might need to be rebased on master.
This looks good.
I am not sure why you did change lots of others tests (by using postcss async api, which make code harder to read).
Also tell me if I am wrong, but with your modification, we are not far from #41 ?

I am just afraid that with the change in compareFixtures function, you didn't update all t.end() called after the call to that function which might miss some unfinished tests.

index.js
jedmao: agreed

test/index.js
jedmao: will do

jedmao:
@MoOx maybe I can decompose this work into 3 separate PRs:
- The plugins option
- The postcss async api
- Asynchronous file I/O #41

Stay tuned.

MoOx:
cool. Can't wait for those things :)

jedmao:
@MoOx I've updated this PR with the following things:

1. Rebased on master.
1. Found one `t.end()` that wasn't being called properly and fixed it. I also made use of tape's `t.plan(n)` method to ensure the number of tests we expect to run are actually being run. Currently, master is running 25 tests in total. I've added 3 tests for a new total of 28, so I'm sure nothing is left behind.
1. Removed the real plugins and replaced with stupid plugins.
1. Throw an error when plugins option is not an array.
1. Squashed changes.

I want to explain why I changed this to use the async API. Actually, I tried keeping things sync at first with `postcss().process(newStyles).css`, but I got errors because some of the plugins in my plugins option were asynchronous; namely, [postcss-bem-linter](https://github.com/postcss/postcss-bem-linter/blob/333f84fd0336b457793ec7d6a817adb87fdd6bb1/index.js#L40). Accepting this fact, I went async. Things got a little weird at that point and the tests were throwing a lot of errors, which I was able to resolve by making this entire plugin async. Of course, this means the tests also had to be modified to be async in order to work.

You're right that we're closer to #41 with this PR.

jedmao:
Also, I tried decomposing this PR into two separate PRs, but I couldn't get it to work without this plugin being async. That's why it's all still in the same PR.

jedmao:
@MoOx you might also find this PR easier to review by [removing whitespace](https://github.com/postcss/postcss-import/pull/55/files?w=1).

MoOx:
You did a great job here ! (I know the whitespace trick, thanks ;)

I just see one last detail https://github.com/postcss/postcss/blob/master/docs/api.md#resultwarntext-opts
The api expect some text, but a warning returned by warnings() is an object. So maybe we should just concat the `result.messages` array.

I will do that and add proper test to ensure messages are ok :)

MoOx:
Released as 6.1.0

jedmao:
Your release notes say it's a `plugin` option in the singular, but it's `plugins` in the plural.

MoOx:
Oh my bad, I will fix that :) Thanks!

PR #58, opened 2015-06-19
Fix plugins section of README

jedmao:
The `Default` part to the plugins option was on the same line as the Type. Simple fix here.


MoOx:
thx

PR #59, opened 2015-06-20
Refactor index.js

### LINKED ISSUE #49: Refactor (explode) index.js
This file is becoming huge. All functions accept too many parameters, it's going to be a pain to maintain if we don't do something :)

jedmao:
Fixes #49 

I got rid of most of the parameters by moving some function declarations into the `AtImport` function block.

### DIFF
--- a/index.js
+++ b/index.js
@@ -32,20 +32,20 @@ var warnNodesMessage =
 /**
  * Inline `@import`ed files
  *
- * @param {Object} options
+ * @param {Object} pluginOptions
  */
-function AtImport(options) {
-  options = assign({}, options || {})
-  options.root = options.root || process.cwd()
-  options.path = (
+function AtImport(pluginOptions) {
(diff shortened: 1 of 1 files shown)


MoOx:
Yeah that cool. I will try to get #47 merged before this one to make the merge easier. (I will be able to do that on tuesday, I am off for the weekend)

jedmao:
No rush on this one.

jedmao:
@MoOx this has been rebased. Looks like #47 needs some more work?

MoOx:
I don't think moving some function declarations in others functions are a good idea, because this will make even more difficult a potential refactoring in several files (right now, this plugin is hard to maintain because of this). At the start it was a pretty small files, and I accepted more and more PR, actions which lead to the current state of the code base.
Instead of just moving some code to avoid some headers, we should find a better way to explode this file into several pieces.

Btw, like an all projects I currently work on, I would go for using babel, or at least some es6/import stuff.

Issue #67, opened 2015-07-09
File containing same content shouldn't be skipped


hgl:
Currently, if a file contains the same content as another, it's skipped. This can fail in cases like this:

```css
/* main.css */
@import "foo/index.css";
@import "bar/index.css";
```

```css
/* both foo/index.css and bar/index.css has this content  */
@import "baz/index.css";
```

This causes `bar/baz/index.css` not being imported.

hgl:
I'm not sure I understand the rationale behind hashing a file, so not sure how to send a PR.

Shouldn't files be skipped only according to their absolute paths? Maybe I missed something?

MoOx:
"Proxy" files (which contain only some `@import`) should be imported all the times, never kipped.

MoOx:
But if full paths are exactly the same, the file will be imported once.

hgl:
Well, since the title fits what I'm about to say. I will just post here.

Imagine this css set up:

```css
@import "tabs/base/index.css";
@import "tabs/pretty/index.css";
```

Both `tabs/base/index.css` and `tabs/pretty/index.css`:
```css
@import "color.css";
@import "background.css";
@import "typography.css";
```

Both `tabs/base/background.css` and `tabs/pretty/background.css`:
```css
.tabs-tab {
    background: url(tab-bg.png)
}

.tabs-panel {
    background: url(panel-bg.png)
}
```

In this case, the base module provides basic styles, and the pretty module provides more vibrant styles.

If you skip files based on content, backgrounds offered by the pretty module are ignored.

hgl:
Probably not skipping files with same absolute url is too much, and multiple modules importing the same normalize.css file is understandable, but files with different absolute urls and same content should not be skipped, i believe.

MoOx:
In order to prevent use case like you mentionned, I was thinking we could (in addition to the content, add a minimal length (in characters) where we can consider files are likely to be the same. But since this is starting to be ridiculous, I guess we should just add an option to enable/disable skipping like it is atm. PR is welcome for a `skipDuplicates` option (like I said in the other PR, true by default to release this as a minor change _for now_).

Reminder: this "feature" is out for months and you are the first to complain about it. I am going to add some documentation about that in order to make it clear for everyone. Your PR (if you are going to do it) might helps others like you. If some others people come to yell, we might change default behavior in a major release.

hgl:
[...]

PR #68, opened 2015-07-09
Add failing test for same file contents

hgl:
I can confirm that the file is not imported.


hgl:
PTAL

I manage to come up with a fix. The crux here is files containing _any_ import rules should not be skipped.

MoOx:
From what I understand you are getting some files that looks the same with import(s) + rules but imports refer to different files ?

hgl:
Yep, exactly.

hgl:
Also I now think skipping files silently might not be a good idea. If can cause some rule failed to be overridden.

For example

```css
@import "foo/red.css";
@import "reset.css";
@import "bar/red.css";
```
and both `foo/red.css` and `bar/red.css` have this content:

```css
body {
    background: url(red.png);
}
```

reset.css
```css
body {
    background: none;
}
```

If `bar/red.css` is skipped, there will be no background on body. And even if the two `red.css` file have the same file, they refer to different images, so really shouldn't be skipped.

I think skipping files based on absolutely url is good enough.

MoOx:
Good point. Any objection @magsout ?

hgl:
Well, just realize skipping files based on absolutely url is wrong too:

```css
@import "foo/red.css";
@import "reset.css";
@import "foo/red.css";
```

I think for css, no files should be skipped.

magsout:
>Also I now think skipping files silently might not be a good idea. It can cause some rule failed to be overridden.

I understand that the test on the hash can introduce a bug on your example. Even though I do not consider it as a real bug. But I try to understand the relevance of your example. It's not really logical to do that, if I may? Or at compile we can specifie that the file is skipped because the same content as such a file.

hgl:
@magsout, you can not expect users to be always logical and not make mistakes like this.

I think we should leave the job of removing unnecessary rules to a css compressor.

MoOx:
I agree that @hgl example doesn't really make sense in practise.
Maybe we should add an option for this, so people can choose to enable or disable skipping.

@hgl the idea was (for example) to avoid having normalize.css several times if some modules use it.

And importing normalize (still for the example) can override previously defined values... So in both cases, so "bugs" are possible...

MoOx:
@hgl you are the first to complain about this "feature", so I guess for now, we should just add an option to allow changing this behavior. What bout `skipDuplicates` (default to true, in order to add this as a minor change). PR welcome :)
Seems good to you @magsout ?

magsout:
>What bout skipDuplicates (default to true, in order to add this as a minor change). PR welcome :)


LGTM :+1:

hgl:
@MoOx what we are currently discussing is if files with the same contents (but contain no imports) should be skipped. But this PR is to not skipping files containing imports (the previous implementation seems to also do that, but with a plain `/@import (.*);/` replace, it can fail in many cases).

I hope this PR is still good for merging.

MoOx:
Yeap.

hgl:
I still think skipping files with same content is a bad idea. I will open a new issue with a more practical example.

MoOx:
https://github.com/postcss/postcss-import/issues/67 ?

hgl:
Yep, I will add to that issue. writing it.

PR #73, opened 2015-08-10
Update dependencies

TrySound:
es6-promise now ignores tests. So, package is much smaller.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -18,6 +18,8 @@ rules:
   comma-dangle: [2, "always-multiline"]
   comma-style: [2, "last"]
   computed-property-spacing: [2, "never"]
+  object-curly-spacing: [2, "never"]
+  array-bracket-spacing: [2, "never"]
   dot-location: [2, "property"]
 
   one-var: [2, "never"]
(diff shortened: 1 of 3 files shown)


.eslintrc
```diff
@@ -9,7 +9,7 @@ env:
 
 rules:
   indent: [2, 2] # 2 spaces indentation
-  max-len: [2, 80, 4]
+  max-len: [2, 100, 4]
```
MoOx: Don't change a rule without asking. This is not the scope of the PR. Tahnks.

TrySound: That's why you need to confirm PR, not just merge. It's asking.
eslint doesn't like this in test/index.js. That string has 100 chars.
```
(t,
arg,
arg)
```

TrySound:
Does it better?

.eslintrc
MoOx: screen shot 2015-08-11 at 10 03 56
Not sure why you needed to change this...

TrySound: There was a few extra spaces.

PR #74, opened 2015-08-18
add options.cache to save file io

zoubin:
I'm using postcss to split css into modules, and postcss-import plays a great role. As there are increasingly many css files to generate, some common modules imported may have to be read many times, which makes the performance worse and worse.

Suppose there are two files to compile:

`a.css`

```css
@import "mixins.css";
.a{}
```

`b.css`

```css
@import "mixins.css";
.b{}
```

It seems that `mixins.css` will be read twice.

I think `options.cache` is able to fix that problem, by caching the transformed file contents and sharing it globally.

Moreover, if the `cache` is manipulatable externally, some file watching tools (like [watchify](https://www.npmjs.com/package/watchify) ) can also work with postcss easily.

I see https://github.com/postcss/postcss-import/pull/47 handles in some way. But that pr has failing checks.

What are your opinions?

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,6 +160,13 @@ It's to optimize output and skip similar files like `normalize.css` for example.
 If this behavior is not what you want, just set this option to `false` to
 disable it.
 
+#### `cache`
+
+Type: `Object`  
+Default: `null`
+
+Cache to save expensive file io. It maps the resolved file name to the corresponding transformed file contents (`String`).
+
 #### Example with some options
 
 ```js
--- a/index.js
+++ b/index.js
@@ -74,6 +74,7 @@ function AtImport(options) {
       importedFiles: {},
       ignoredAtRules: [],
       hashFiles: {},
+      cache: options.cache || {},
     }
     if (opts.from) {
       state.importedFiles[opts.from] = {
@@ -347,7 +348,8 @@ function readImportedContent(
     options.encoding,
     options.transform || function(value) {
       return value
-    }
+    },
+    state.cache
   )
 
   if (fileContent.trim() === "") {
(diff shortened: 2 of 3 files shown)


MoOx:
I guess it's a nice start & a good small step to an even more smart cache (parsed ast in a huge file like #47). Tests looks good.
I think we should merge this one so we can iterate more easily. What do you think @
andyjansson ?

zoubin:
>I guess it's a nice start & a good small step to an even more smart cache (parsed ast in a huge file like #47). Tests looks good.
I think we should merge this one so we can iterate more easily. What do you think @
andyjansson ?

@andyjansson

andyjansson:
@MoOx  @zoubin So, judging by the code, this is to inject CSS via JavaScript..? That doesn't seem like a cache to me. 

Also, speaking of #47, everything is implemented. The only reason it has failing checks is because it's out of sync with the upstream repo - the tests pass!  I just need help bringing it up to speed with the upstream repository. @jedmao did a PR which refactored the code base which might things a bit more difficult to merge, but I think the code is overall better for it. 

So, if anyone is able to help, I'd be grateful.

zoubin:
@andyjansson 
>So, judging by the code, this is to inject CSS via JavaScript..? That doesn't seem like a cache to me.

Perhaps the test code is a little misleading.
I do not mean to inject CSS via JavaScript. The contents of `options.cache` are only set by postcss-import. When postcss is run multiple times, this cache will persist, so that each file will be read at most one time.

Another reason for `options.cache` is that we can watch imported files given by `onImport`, and if any watched file changes, we can invalidate the corresponding cache in `options.cache`, and restart the compilation to renew the result.

andyjansson:
@zoubin This is already handled in #47. It caches the file, reads it once and invalidates the cache automatically if the file has been modified.

andyjansson:
In #47, this is what happens to the cache if a modification to a file in the dependency graph is detected:

Issue #39, opened 2015-04-27
Can we add the option to import underscored file names like Sass


Designer023:
In Sass you can import files with underscores without adding underscores in the code- When the file name is actually _my-file.scss :

    @import "my-file.scss"
    // or 
    @import "my-file"

Unless I'm missing something it would be handy to add this. Having loads of underscores in the names makes it ugly to read and its also useful to have a prefix that's ignored so we can exclude the files from watch tasks etc

Thanks you

MoOx:
Hum, not sure I want this sassy thing in here.
That's specific to sass & his way to build non _* files & consider _* files as partial (not to build as output files).

bloodyowl:
maybe let people implement a custom path resolver could do without making it sass-specific.

MoOx:
good idea indeed @bloodyowl

srcspider:
+1 to simple path resolver

Would personally like to be able to specify files relative to a root.

MoOx:
> Would personally like to be able to specify files relative to a root.

You have [`root` and `paths` options](https://github.com/postcss/postcss-import#options) for that

SlexAxton:
Might be what you're looking for https://github.com/jonathantneal/postcss-partial-import

TrySound:
Ref #145

MoOx:
Closing in favor of #145

PR #79, opened 2015-09-07
Suport for alternative PostCSS syntaxes

jonathantneal:
Added: support for alternative PostCSS syntaxes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# 7.1.0 - 2015-09-07
+
+- Added: support for alternative PostCSS syntaxes
+([#79](https://github.com/postcss/postcss-import/pull/79))
+
 # 7.0.0 - 2015-08-25
 
 - Removed: compatibility with postcss v4.x
--- a/index.js
+++ b/index.js
@@ -390,7 +390,9 @@ function readImportedContent(
 
   if (options.async) {
     return parsedResult.then(function() {
-      return processor.process(newStyles)
+      return processor.process(newStyles, {
+        syntax: result.opts.syntax,
+      })
         .then(function(newResult) {
           result.messages = result.messages.concat(newResult.messages)
         })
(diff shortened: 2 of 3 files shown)


MoOx:
I don't understand what this is about. Care to add a test?

ai:
PostCSS 6.0 has custom parsers. So user can work not only with CSS sources, but also with SCSS (for eample, for inline comment). So if it import some file, we need to use parent processor parser.

MoOx:
I see you added postcss-scss as a dev dep but I don't see any tests, did you forget to add them ?

tinchoz49:
:+1: what is the status of this feature? I preffer use postcss-import over to postcss-partial-import but for now i don't have alternative if i'm using SCSS

MoOx:
I am just waiting for a minimal automated test to be able to merge.

tinchoz49:
ahh ok! one more thing. @jonathantneal I think that you need to do some change with the line [367](https://github.com/postcss/postcss-import/blob/master/index.js#L367) because for example for inline comments in SCSS is necessary use postcss-scss as the parser.

maybe something like this, what do you think?
```javascript
var parseFunc = (result.opts && result.opts.parser) 
  ?  result.opts.parser
  :  postcss.parse
var newStyles = parseFunc(fileContent, options)
```

jonathantneal:
Maybe it would be easier if we just merge the two projects, and keep the import functionality like it is in PostCSS import and make the include functionality like it is in PostCSS partial import?

MoOx:
The idea of your plugin is just to add a prefix before the last part of the name of the file right? I don't see how you created another plugin for that and not just did a PR in the first place. This has already been discussed here https://github.com/postcss/postcss-import/issues/39 and the issue is still open (so PR are welcome, has always).

CoolGoose:
+1 on this, any chance of getting this trough ? :) (helps with a lot of projects being sass based)

MoOx:
@CoolGoose If I am correct, I think @TrySound will handle that very soon.

MoOx:
Closed by 9c168b90afee7673df7c617d08a06a076500a499, will land in master soon.

PR #87, opened 2015-10-07
Update README.md to reflect that node modules should set package.json…

oleersoy:
…'s style attribute to the name of the CSS entry file


MoOx:
Thanks !

PR #94, opened 2015-11-05
Simplify promises

TrySound:


MoOx:
Nice catch. I really need to make better review :(

PR #95, opened 2015-11-06
Correct normalize.css bower reference

SevereOverfl0w:
normalize-css seems to have become the preferred directory for both versions of the package, tripped me up initially trying to follow the example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ Using this `input.css`:
 /* can consume `node_modules`, `web_modules`, `bower_components` or local modules */
 @import "cssrecipes-defaults"; /* == @import "./node_modules/cssrecipes-defaults/index.css"; */
 
-@import "normalize.css/normalize"; /* == @import "./bower_components/normalize.css/normalize.css"; */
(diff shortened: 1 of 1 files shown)


TrySound:
:+1:

MoOx:
The "official" npm package is "normalize.css" (https://www.npmjs.com/package/normalize.css) and not normalize-css https://www.npmjs.com/package/normalize-css.
Since bower is kind of deprecated, I will just remove bower from the example since I consider it to be a bad practise.

SevereOverfl0w:
Oops! looks like I messed up. You're quite right.

TrySound:
@MoOx Will you remove bower in 8.0?

MoOx:
No, bower support is just one line of code (one item in an array), I am just removing it from examples since I do not recommend people to use bower (so I think it's better to avoid in examples).

SevereOverfl0w:
@MoOx Any reason why?

MoOx:
Why do you need bower when npm can do the same job (better?)

SevereOverfl0w:
I definitely agree there!

Unfortunately, there's still a few libs only on bower, they're getting fewer though.

PR #99, opened 2015-11-17
Support *.pcss extension

meritt:
See https://twitter.com/PostCSS/status/661645290622083073


TrySound:
:+1:

MoOx:
@meritt just click again and read the whole thread. I think it's a stupid idea and I think we should really not promote that, it's a non-sense. Does babel user use .bjs? I guess no.
Howhever, I am option to add a `extensions` option.

meritt:
@MoOx `extensions` option would be fine for me. You can close PR.

MoOx:
Feel free to make a PR to add `extensions` option.

PR #103, opened 2015-12-13
Simplify root node path

TrySound:
Any reason to deny it?


MoOx:
Back in time this was not possible (from my memory), but I guess this is relevant now.

PR #104, opened 2015-12-14
Remove node using postcss api

TrySound:


TrySound:
https://github.com/postcss/postcss-import/blob/master/test/fixtures/ignore.expected.css
ignore.actual.css
```css
@import "http://css" (min-width: 25em);

@import "http://css-screen" (min-width: 25em) and screen;

@import "http://css";

@import "https://css";

@import 'http://css';

@import 'https://css';

@import url(http://css);

@import url(https://css);

@import url("http://css");

@import url("https://css");

@import url('http://css');

@import url('https://css');

@import url("//css");

@import url('//css');

@import url(//css);

@media (min-width: 25em){
ignore{}
}

content{}
```

MoOx:
It doesn't make sense to have to edit the expected result... Here postcss is still broken....

MoOx:
I don't think the codebase is complicated because of this trick that is required to keep a code as it is expected.

TrySound:
@MoOx This stuff just complicate code. And your tricks not always do what are expected. For example [here](https://github.com/postcss/postcss-import/blob/master/test/fixtures/duplicates.expected.css)

Selectors are expected to be indented
```css
@media screen{
    foo{}
}
```

but we have some flat uncomfortable code.

Yes, maybe normalize do not do everything what you expect, but it do good work which shouldn't do plugins.

Let's simplify codebase with minimal changes in expected code and everything will be okay. :)

MoOx:
[Previously I did that because normalize was bringing bugs](https://github.com/postcss/postcss-import/commit/e985ca694a51ce88747b4e724de3bdcbc495c857)
If you can confirm that postcss have fixed those issues then we might try...

PR #105, opened 2015-12-15
Simplify rules insertion

TrySound:


MoOx:
The previous PR and this basically revert the thing I introduced to fix postcss bugs back in time.
I guess it's better now :)

PR #106, opened 2015-12-15
Use async mode only

TrySound:
For now postcss-import isn't async at all. It still uses sync methods. In the next PR I will fix this too.

### DIFF
--- a/index.js
+++ b/index.js
@@ -31,7 +31,6 @@ var warnNodesMessage =
 function AtImport(options) {
   options = assign({
     root: process.cwd(),
-    async: false,
     path: [],
     skipDuplicates: true,
   }, options || {})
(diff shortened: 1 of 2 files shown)


MoOx:
This was for legacy purpose (cssnext is still using sync api, I am starting to cleaning things in order to deprecate it, so this won't be a problem soon).
This will need to be released as a breaking change. Can you update README (remove async doc) and prepare CHANGELOG accordingly?
If we you want to continue to iterate this way (nice approach) maybe we should merge in a "dev" branch, then merge in master when releasing? Or do you want to ship this as it flows?

TrySound:
@MoOx Okay let's create dev branch.

MoOx:
I created it from master, can you adjust the scope of the PR?

MoOx:
You could not just update the scope? I didn't find a way to do it myself but I was thinking that's because I didn't create it.

TrySound:
@MoOx No, don't have an idea how to do that.

MoOx:
I tried on some opened PRs I got, it seems it's not doable anymore from github ui...

PR #109, opened 2015-12-15
Simplify debug

TrySound:


MoOx:
I think using `/* eslint-disable no-console*/` is a better idea.

TrySound:
@MoOx Okay. Maybe replace tape with `blue-tape` which is support returning promises?

MoOx:
what about ava ? :)

PR #110, opened 2015-12-15
Simplify parseGlob

TrySound:


MoOx:
You are adding more code than you remove but why not :D

PR #115, opened 2015-12-20
Fix multiple media

TrySound:


lib/parse-imports.js
```diff
@@ -1,6 +1,22 @@
 var valueParser = require("postcss-value-parser")
 var stringify = valueParser.stringify
 
+function getMedia(params) {
+  var media = []
+  var last = params.reduce(function(medium, node, index) {
+    if (index < 2) {
+      return ""
+    }
+    if (node.type === "div" && node.value === ",") {
```
MoOx: Why is the a test on a "div" ? Not sure to follow here.

TrySound: div is divider (',', '/', ':')
This function just split params with comma divider after second (url+space) node

TrySound: The reason is
`(cond1 + cond2) * parentCond = cond1 * parentCond + cond2 * parentCond`

MoOx: Wow, "div" is confusing.

MoOx:
Good job.

PR #116, opened 2015-12-21
Refactor path resolve

TrySound:


test/import.js
```diff
@@ -54,12 +55,6 @@ test("should not need `path` option if `source` option has been passed", t => {
   })
 })
 
-test("should be able to consume npm package or local modules", t => {
-  return compareFixtures(t, "modules", {
-    root: ".",
-  })
-})
-
```
MoOx: Why did you remove this test(s)?
Looks pretty important to me.

TrySound:
@MoOx What is the case of local npm modules?

MoOx:
web_modules with webpack :)

MoOx:
Note that I think this also cover node_modules.

TrySound:
@MoOx package.json in node_modules or web_modules still works. But it doesn't work in local directories. Also I removed bower_components, caz we still do not support bower.json

TrySound:
@MoOx Okay I saved previous way to resolve paths.

index.js
```diff
@@ -18,7 +18,6 @@ var resolveMedia = require("./lib/resolve-media")
 var moduleDirectories = [
   "web_modules",
   "node_modules",
-  "bower_components",
```
MoOx: This should be added in the CHANGELOG since it's a breaking changes (and we should add how to add bower support like it was).

test/fixtures/ignore.expected.css
```diff
@@ -28,6 +25,10 @@
 
 @import url(//css);
 
+@import "http://css" (min-width: 25em);
+
+@import "http://css-screen" (min-width: 25em) and screen;
+
```
MoOx: Why is the order changed here? This looks like a total weird side effect that will break user expected result...
If the ignored file is being imported first, imported content should be imported at this place, not later.
Looks like we are introducing a bug :)

test/import.js
```diff
@@ -155,14 +154,16 @@ test("should work with no styles without throwing an error", t => {
...
   const resolve = require("resolve")
-  const sassResolve = (file, opts) => {
-    opts = opts || {}
-    opts.extensions = [ ".scss", ".css" ]
-    opts.packageFilter = pkg => {
-      pkg.main = pkg.sass || pkg.style || "index"
-      return pkg
-    }
-    return resolve.sync(file, opts)
+  const sassResolve = (id, base, opts) => {
```
MoOx: if the signature of the callback to provide custom resolve, we should also document that and specify this in the CHANGELOG.

TrySound:
@MoOx Done!

TrySound:
Is PR too big or you can accept it?

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 - Removed: async mode/option
 ([#107](https://github.com/postcss/postcss-import/pull/107))
+- Removed: bower is not currently supported, use custom resolver
+([#116](https://github.com/postcss/postcss-import/pull/116))
```
MoOx: fyi, you can just add bower support by it to the `path` option.
So I would use ``"bower_components" not supported by default anymore, use "path" option to add it back``.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
...
+- Changed: custom resolve has more responsibility for paths resolving
```
MoOx: This message is not clear. We should explicitly tell what has changed in the signature of the function (or send the user to the right place in the doc).

CHANGELOG.md
TrySound: bower support is also bower.json support, which is not simple with main field as array.

MoOx:
@TrySound Except my last comments about the CHANGELOG, this PR looks good. Great job. I just want to be sure that the developer experience for upgrading is easy :D

CHANGELOG.md
TrySound: Not it just gets `uri` and `baseDir` and should return resolved path.

CHANGELOG.md
MoOx: I know that, but bower is more a downloader, so I guess lots of people just point to the right css file when they import it (eg: `@import "jquery-slider-fx-2016/slider.css"`) so we don't really care about it. The idea is just to inform people that are currently using the feature we are breaking how to upgrade easily.

CHANGELOG.md
MoOx: So you can add at the end of the messsage: "See XXX for more information about this change" to point to the right place in the README.

TrySound:
@MoOx Done

MoOx:
:+1:

PR #119, opened 2015-12-24
Simplify functions: less arguments, code refactoring

TrySound:


MoOx:
less code added than removed + no change in tests: lgtm :)

PR #120, opened 2015-12-24
Add array of resolved files support

TrySound:
A bit more code but I'd like to move glob responsibility on resolve-id


MoOx:
Nice.

PR #123, opened 2016-01-02
Use eslintrc for simple package.json updating

TrySound:
I can't do `npm i something -D` without changing default package.json codestyle. Also yaml seems much prettier.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,0 +1,33 @@
+extends: "eslint:recommended"
+ecmaFeatures:
+  modules: true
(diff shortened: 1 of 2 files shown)


MoOx:
Or we can just accept npm changes? I like not having rc files these days.

TrySound:
@MoOx It's really big and ugly.
```
"ruleName": [
  "2"
],
...
```

MoOx:
Ok :(
Not a big deal anyway.

PR #125, opened 2016-01-02
Remove unnecessary glob-alt test

TrySound:
resolver response for globs. They use the same parser as regular uris


MoOx:
Good catch.

PR #127, opened 2016-01-03
Simple entry path detecting

TrySound:


TrySound:
opts.from is not documented as I know.

MoOx:
https://github.com/postcss/postcss/blob/master/docs/source-maps.md#postcss-and-source-maps ?

TrySound:
@MoOx I mean it's from option of plugin, not process

MoOx:
Hahaha. Oups.
Maybe we should postcss from option so?

TrySound:
@MoOx source,inout.file is postcss.from

MoOx:
True :)

PR #128, opened 2016-01-03
Reduce options assigning

TrySound:
Nothing is changed there

### DIFF
--- a/index.js
+++ b/index.js
@@ -20,7 +20,7 @@ function AtImport(options) {
     root: process.cwd(),
     path: [],
     skipDuplicates: true,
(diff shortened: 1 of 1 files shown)


MoOx:
Not sure. This may introduce mutability issues (not currently covered by any tests tho).
But we will see :)

TrySound:
@MoOx There are some mutations deeper, but assign is used there.

MoOx:
That might create issue between plugin runs with the same instantiation.
But not test: no bug (lol). If the issue came up, we will fix+ cover by tests.

PR #131, opened 2016-01-04
Simple glob resolver

TrySound:
`moduleDirectories` shound't be used caz they do not give any resolving benefits and can just confuse. Better give ability to add it to path.
Also all resolvings should be consistent with enabled `glob`, so non-glob imports should also be resolved as glob.


MoOx:
This should be documented.

TrySound:
@MoOx Is it okay now?

MoOx:
I am double checking this and do not really understand "why is this confusing to have node_modules and web_modules checked by default - like non globbed imports?"

TrySound:
@MoOx Because they just paths. They won't be resolved as modules with package.json or index.css. In this case we should prevent some implicit behaviour.

MoOx:
Well... Not sure that a consistent choice. Anyway if you feel it is, why not, I never used globbed import.

PR #135, opened 2016-01-04
Prepare parse statements

TrySound:


test/lint.js
```diff
@@ -4,23 +4,6 @@ import atImport from ".."
...
-      const warnings = result.warnings()
-      t.is(warnings.length, 1)
-      t.is(
-        warnings[0].text,
-        "@import should not be inside block statement"
-      )
-    })
-  })
-})
-
```
MoOx: This wasn't that good ?

TrySound: I'd like to improve performance here. We don't need to traverse all tree. And imports will be just leaved in block statement.

PR #136, opened 2016-01-06
Statements

TrySound:
Now parseStatements returns also `nodes` and `media` types.
In parseStyles I just filter them out. I will use and test them in the next PR.

Everything to make correct async behaviour.


MoOx:
Nice :)

PR #137, opened 2016-01-06
Refactoring

TrySound:
This is the better decision. We should process imports only after plugins transformations.

### DIFF
--- a/index.js
+++ b/index.js
@@ -93,25 +93,24 @@ function parseStyles(
   processor
 ) {
   var statements = parseStatements(result, styles)
-  var importResults = statements.filter(function(stmt) {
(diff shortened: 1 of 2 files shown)


MoOx:
> We should process imports only after plugins transformations.

This point is not clear to me. Can you elaborate a little bit?

TrySound:
@MoOx Maybe in plugins some imports will be removed. In current case it's the same as passing plugins in the next pipe. But with my change plugins will process source files and then they will be process by postcss-import

MoOx:
Not sure if I understand this correctly but it seems like a step backward.
What if a var is declared in a imported file and used after?

PR #139, opened 2016-01-07
Process media joining in parseStyles

TrySound:


MoOx:
Maybe we should cut a major release and see if there is not too many issues ?

TrySound:
@MoOx I'd like to fix media nesting and then publish major release.

PR #140, opened 2016-01-08
Use statements manipulation

TrySound:
I know this pr is very big but I can't make it smaller :)


MoOx:
Can you elaborate a little bit? Cause it does not remove or simplify the code at the first look :)

TrySound:
@MoOx This pr simplify the process. Okay.
This changes let use to manipulate small chunks/statements (`import`, `media`, `nodes`)
This will let to make good media nesting according spec.
I will add some tests for them in the next PR, but I think they will be passed.

Issue #133, opened 2016-01-04
spec following


TrySound:
Should we follow [spec](http://www.w3.org/TR/css3-cascade/#at-import) and skip and warn if import statement is not before all non-import rules?
This will let resolve media queries much faster and simpler. For now resolver do not follow expectations and it's very difficult to follow them without spec.

/cc @MoOx @ai @ben-eb

ben-eb:
Sounds good to me. :+1:

TrySound:
But we won't be able to do something like this
```
@import "normalize.css";
body {
 margin: 20px;
}
@import "module.css"
```

TrySound:
For now the problem is nesting

```css
/* level-1.css */
@import "level-2.css" level-1;
@import "//level-1";
```
```css
/* level-2.css */
@import "level-3.css" level-2;
@import "//level-2";
```
```css
/* level-3.css */
@import "//level-3";
```
what expected
```css
@import "//level-3" level-1 and level-2;
@import "//level-2" level-1;
@import "//level-1";
```
reality
```css
@import "//level-3" level-2;
@import "//level-2" level-1;
@import "//level-1";
```
Also media rules are not resolved.

MoOx:
Would be closed automatically when your pr/commit land into master.

PR #142, opened 2016-01-09
Add media queries tests

TrySound:
New tests works perfectly. This was not possible with old structure. Now we can fix problems fast and furious. :))


TrySound:
@MoOx  Can you add some impressive message for this change? :)

TrySound:
This was before
https://github.com/postcss/postcss-import/issues/133#issuecomment-168794761

TrySound:
By the way also it process media-queries statement now ;)

MoOx:
Nice work.

PR #143, opened 2016-01-09
Refactor resolve option

TrySound:


MoOx:
Maybe we should add in the doc something for the end user that want to consume files a la Sass?

MoOx:
Need a rebase

TrySound:
@MoOx We can just add example in documentation after option definition

MoOx:
Yeah.

PR #144, opened 2016-01-09
Add custom load support

TrySound:


TrySound:
@MoOx Maybe we should remove glob support? I created api which let in fast way add this feature.
```
import({
  resolve: function (id, base, opts) {
    return glob.sync(path.join(base, id));
  }
})
```

MoOx:
I am sure a lot of people use glob and don't want to type this function. Let's keep it in the core.
Why did you removed encoding option?

TrySound:
@MoOx Cause I replaced it with custom load. I think it more consistent way.

MoOx:
So please can you add in the CHANGELOG a way to migrate? Thanks :)

TrySound:
Like this?

TrySound:
I just would like to keep core as small as it possible. And glob support is weird in browser way. We made good and simple api for that. glob can be realized in the same way as encoding support. Also globs can be realized in different ways. What we did can be not what users expect.

MoOx:
So if we remove things like that, I really think we should provide snippets "ready to use" to provide an easy upgrade.
Can you handle those? They should be in the CHANGELOG.

TrySound:
@MoOx The same like encoding example?

MoOx:
Yeah that's cool. (btw, why did you used fs sync in it?)

TrySound:
@MoOx It's simpler than wrapping async in promise :)

PR #148, opened 2016-01-09
Fix tests using the same fixtures

TrySound:


MoOx:
What was the issue?

TrySound:
I use here the same fixtures for different tests. I'm not sure but can be some conflict.

MoOx:
Ok.

Issue #48, opened 2015-06-04
Add caching to improve performance


andyjansson:

MoOx:
Closing in favor of #47

PR #152, opened 2016-01-11
Add read-cache to load files

TrySound:


MoOx:
Is this closing #47 #48 #74 ?

MoOx:
(You can provide a small description on your PR :p)

TrySound:
I think it solve those issues only partially. read-cache just cache readFile result checking mtime. But it do not cache transformations. Let's leave those issues for the future.

MoOx:
ok. Good start. :)

PR #165, opened 2016-02-02
Add bower_components to module directories

scotthovestadt:
We're using this to pull in Bootstrap CSS from bower_components. Can we add this as a directory to automatically search?

[Bower](http://bower.io/) is a relatively popular package manager.

Thank you for your time!


scotthovestadt:
I'm adding tests. My mistake.

scotthovestadt:
I see that you had bower support previously and removed it. Closing.

PR #170, opened 2016-02-13
Change docs of `addDependencyTo` option

git-in-my-anus:
Fix type declaration and description. Provide usage example with postcss-loader


MoOx:
Thanks !

Issue #171, opened 2016-02-13
Master is broken because of ava/babel@6.5.1


MoOx:
https://github.com/sindresorhus/ava/issues/538

MoOx:
Fixed.

PR #172, opened 2016-02-13
test: fix failing test due to dependency

mastilver:
closes: #171


MoOx:
Thanks but that's ok now.

PR #173, opened 2016-02-16
Add clarification on Promise usage in transform

simonsmith:
I got a bit confused and tried to return just a Promise before checking the tests and realising it expects a resolved Promise:

``` js
transform(css) {
  return postcss([aPlugin]).process(css); // didn't work
}
```

I've added a note to make it clearer

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,8 +126,16 @@ Type: `Function`
 Default: `null`
 
 A function to transform the content of imported files. Take one argument (file
-  content) and should return the modified content or promise with it.
+  content) and should return the modified content or a resolved promise with it.
 `undefined` result will be skipped.
+
+```js
+transform: function(css) {
+  return postcss([somePlugin]).process(css).then(function(result) {
+    return result.css;
+  });
+}
+```
 
 #### `plugins`
 


TrySound:
@simonsmith Exactly. Thanks. Did you see new recipe? You can just pass plugins in postcss-import plugins option.
https://github.com/stylelint/stylelint/blob/master/docs/user-guide/postcss-plugin.md#usage-examples

TrySound:
Thanks for your idea.

MoOx:
It's more that transform expect some CSS, not a postcss "result", don't you think?

simonsmith:
> You can just pass plugins in postcss-import plugins option.

Yeah, that's a great feature but in my use case I need a little bit of logic (rather than just using the plugin) so that's why I stuck with `transform`

> It's more that transform expect some CSS, not a postcss "result", don't you think?

Sure, it's pretty obvious now but I guess I always think when I read "can return a promise" I don't think about resolving it to a specific value first as the code receiving it could do that for me. And you can never write too much documentation :)

MoOx:
Yep.

Issue #181, opened 2016-03-03
Exposing resolve/load is great, but implementation could use a little work.


MeoMix:
Hey there,

I am trying to use `postcss-import` with JSPM. During development I run this plugin from a browser environment rather than using node.

As such, I provide custom `resolve` and `load` methods (which work great, thanks!):

```
  atImport({
    resolve(id, basedir) {
      // at-import returns a directory name without a trailing slash.
      // The lack of trailing slash causes System.normalize to generate an incorrect URL.
      const canonicalParent = `${basedir.replace(/^\//, '')}/`;
      return System.normalize(id, `${System.baseURL}${canonicalParent}`);
    },
    load(filename) {
      return System.import(filename);
    }
  }),
```

However, I encountered one other issue which I needed to fork the project to resolve.

By default, `postcss-import` uses the `load-content` module as the default value for `load`. `load-content` has a dependency on `read-cache`. `read-cache` attempts to pass `fs` methods to `pify` which errors out due to being in a browser environment.

This is frustrating because I'm overriding the `load-content` dependency with my own `load` method, but, as the dependency is eagerly loaded, `postcss-import` is still coupled to node environments.

Do you think there's an elegant way to remove this coupling?

TrySound:
@MeoMix You can replace require with undefined in your builder.

MeoMix:
@TrySound Mmm, I think I know what you mean. Good point. I'll fuss with it a bit and report back.

MeoMix:
Got it! Thanks for pointing me in that direction.

Solution was to add a custom map configuration into jspm:

```
    "npm:postcss-import@8.0.2": {
      "map": {
        "object-assign": "npm:object-assign@4.0.1",
        "postcss": "npm:postcss@5.0.19",
        "postcss-value-parser": "npm:postcss-value-parser@3.3.0",
        "read-cache": "npm:read-cache@1.0.0",
        "resolve": "npm:resolve@1.1.7",
        "./lib/load-content": "@empty"
      }
    },
```

PR #185, opened 2016-03-07
Support running PostCSS-import via JSPM in browser environment

MeoMix:
By adding an entry to package.json which is jspm-specific we allow PostCSS-import to be utilized by JSPM in browser environments.

The `load-content` module has node-specific deps. (mentioned here: https://github.com/postcss/postcss-import/issues/181). As such, it should be prevented from being loaded in browser environments in favor of specifying a custom `load` via options.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,5 +39,11 @@
   },
   "eslintConfig": {
     "extends": "eslint-config-i-am-meticulous/es5"
+  },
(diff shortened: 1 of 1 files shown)


TrySound:
@MeoMix I think resolve should be empty too. And squash commits please.

PR #194, opened 2016-04-20
JSPM Lookups

swernerx:
Added support for JSPM based lookups.

I have not yet any tests for jspm lookup in this package, but in the origin/dependency project (would require some jspm installs which I am not sure should be done here).


lib/resolve-id.js
```diff
@@ -23,7 +24,7 @@ module.exports = function(id, base, options) {
     basedir: base,
     moduleDirectory: moduleDirectories,
     paths: paths,
-    extensions: [ ".css" ],
+    extensions: [ ".css", ".sss" ],
```
MoOx: Please don't add this. You can handle this via option.

package.json
```diff
@@ -18,7 +18,9 @@
     "lib"
   ],
   "dependencies": {
+    "jspm": "^0.17.0-beta.13",
```
MoOx: Maybe devdeps?

.eslintignore
```diff
@@ -0,0 +1,3 @@
+jspm.browser.js
```
MoOx: why don't you lint those files?

lib/resolve-id.js
swernerx: Okay, I'll change it back.

package.json
swernerx: That seems right. I'll change it.

.eslintignore
swernerx: Mostly because they are generated/modified when using jspm. They are not really human-controlled.

MoOx: so sshould not be add in git?

swernerx:
I reduced the integration for jspm a bit. This was mainly a try to solve the test puzzle for jspm. The current versions seems to be the minimum viable one to add jspm packages resolving...

package.json
```diff
@@ -28,9 +29,20 @@
...
   },
+  "jspm": {
+    "name": "postcss-import",
+    "main": "index.js",
+    "browser": {
+      "./lib/load-content": "@empty",
+      "./lib/resolve-id": "@empty"
+    },
+    "dependencies": {},
+    "devDependencies": {}
```
MoOx: are the empty field required?

swernerx: You are really caring about the details, are you? - That's welcome.

swernerx: I will try with leaving them out...

swernerx:
Looks good without dependencies and devDependencies.

swernerx:
Would it be possible to merge this into master and do a new release? Or is there anything outstanding we have to work before this could happen?

MoOx:
Released as 8.1.1

jedrichards:
Adding `pkg-resolve` into the dependency tree caused my builds to fail with a ton of security warnings when I just upgraded to the latest `postcss-import`. I know these security warnings aren't necessarily relevant for packages used in build processes as opposed to outward facing webapps, but still thought you might be interested.

screen shot 2016-05-18 at 16 45 31

Also it seems a slight shame that adding jspm support means that all users have the additional complexity overhead of the jspm modules added to their dependency trees, even if they don't use jspm ... hence problems like the above ...

MoOx:
I think we should move this to peerDependencies and only import in a jspm context. Will open an issue about that.

MoOx:
https://github.com/postcss/postcss-import/issues/212

MoOx:
@swernerx any thoughts about the security issues?

aunz:
I use postcss-import in webpack postcss-loader
Addition of JSPM in version 8.1.1 or higher creates this error: `Module build failed: Error: Unable to fetch "C:\XXX\node_modules\babel-loader\index.js". Only file URLs of the form  allowed running in Node.`
A similar [issue](https://github.com/systemjs/systemjs/issues/306) was described in systemjs
Version 8.1.0 and below without JSPM has no such error
I think this happens only in Windows
My system is Windows 10 64 bit with node 6.2.0
Took me a while to figure this out, hope this can be useful for windows users who experience the same issue?

PR #200, opened 2016-05-04
Fixed extension override

swernerx:
Somehow it seems like defining an array of alternative extensions never made it through the import chain. This PR fixes the issue.


swernerx:
Will be replaced

Issue #198, opened 2016-05-04
Issue with 8.1.1 - JSPM Permission denied


nwhite89:
I'm using `postcss-import` with webpack since the update to `8.1.1` and using the `pkg-resolve` module I am getting an error when running

```js
Unable to create jspm system folder
Error: EACCES, permission denied '/home/ubuntu/.jspm'
```

When I am running my project I don't have permissions to this folder and I believe it's unnecessary also I'm not entirely sure why `jspm` is even needed to be installed? It's not even used by `postcss-import`. For now to get this working I have had to downgrade to `8.1.0` any help on this would be greatly appreciated :). 

Thanks.

MoOx:
Sorry for this @nwhite89. I am not even using jspm myself.

This is related to https://github.com/postcss/postcss-import/pull/194 that I accepted.
I guess it's because you ran into an undefined file, which end up in the catch handled by this PR.

@swernerx any idea how we can make the resolve optional? (only used in jpsm env)

nwhite89:
@MoOx thanks for your quick reply what is the reasoning to use `pkg-resolve`? Looking at the code change I can't quite follow the necessity for using the package.

MoOx:
It's to ensure JSPM support (but to be honest, I blindly trusted @swernerx, so we should wait for his input).

swernerx:
Oops. That sounds "interesting". I never saw that issue here. 

Yep, pkg-resolve is for resolving jspm which is a pretty useful package manager for resolving the front-end part of typical webapps. 

I am not sure what happens there on your system that it throws such an error. JSPM seems to try to create the "~/.jspm" folder during runtime. The questions is why this raises an error at all. Shouldn't the app be able to create config files in the home directory?

JSPM uses the folder for user wide "global" caching of dependencies, etc. This is not implemented in "pkg-resolve" which is just querying "jspm" for its package registry.

MoOx:
> EACCES, permission denied

Can be a "read" issue.

Is there a way to detect during runtime that we are in jspm env? Because if not, using pkg-resolve will only leads to error (eg: npm + postcss only). We should avoid that.

swernerx:
It seems like JSPM is pretty intolerant with any unexpected situations regarding this folder. It throws with hard exceptions for any cases were:

- The home directory is not writable
- The directory exists but is not writeable
- There is a file which uses the same name instead of a directory.

I figure for me "jspm" should ignore all of these failed attempts and maybe present a warning instead of crashing.

[...]

PR #205, opened 2016-05-05
Improved tolerancy for failed imports of jspm-chain. See issue #198. Works around https://github.com/jspm/jspm-cli/pull/1779 stuntil that is fixed.

swernerx:
Clean PR of previous fix


MoOx:
Can't we detect first that a user is using jspm (or not) in order to not even require(pkg-resolve) if not in a jspm context?

swernerx:
I do not think that this is super relevant as soon as jspm itself is fixed.

PR #230, opened 2016-09-26
Fix tests

guybedford:
This gets the tests running again, and also updates to the latest version of Ava. The issue seemed to be related to the media property not being set to an empty array by default.


MoOx:
Thanks!

PR #231, opened 2016-09-26
Add input sourceMap support for load hook

guybedford:
This allows the `load` option to return a `{ source, sourceMap}` object like the following:
```javascript
load: () => {
  return {
    source: 'css source',
    sourceMap: {cssSourceMap}
  };
}
```

While retaining backwards compatibility for returning the direct source string as well.


RyanZim:
I like this idea.

@ai @MoOx Thoughts on this?

@guybedford Merge conflicts will need to be fixed; it may be easier to re-implement this instead of rebasing since the codebase has changed quite a bit. Most notably, we have removed the `transform` option, which will greatly simplify this.

ai:
Hm. Looks important. But keys in return object looks non common. Could somebody check other systems API, what keys they use in returning source/map.

PR #236, opened 2016-11-02
Nested import ordering

### LINKED ISSUE #168: Incorrect nested import ordering

RyanZim:
Had to edit a test that depended on the old, incorrect behavior.

Fixes #168 :tada:

### DIFF
--- a/index.js
+++ b/index.js
@@ -5,6 +5,7 @@ var joinMedia = require("./lib/join-media")
 var resolveId = require("./lib/resolve-id")
 var loadContent = require("./lib/load-content")
 var parseStatements = require("./lib/parse-statements")
+var promiseEach = require("promise-each")
 
 function AtImport(options) {
   options = assign({
@@ -163,7 +164,7 @@ function parseStyles(
 ) {
   var statements = parseStatements(result, styles)
 
-  return Promise.all(statements.map(function(stmt) {
+  return Promise.resolve(statements).then(promiseEach(function(stmt) {
     stmt.media = joinMedia(media, stmt.media || [])
 
     // skip protocol base uri (protocol://url) or protocol-relative
--- a/package.json
+++ b/package.json
@@ -22,6 +22,7 @@
     "pkg-resolve": "^0.1.7",
     "postcss": "^5.0.14",
     "postcss-value-parser": "^3.2.3",
+    "promise-each": "^2.2.0",
     "read-cache": "^1.0.0",
     "resolve": "^1.1.7"
   },
--- a/test/order.js
+++ b/test/order.js
@@ -3,27 +3,13 @@ import compareFixtures from "./helpers/compare-fixtures"
 
 test(`should order nested imports correctly`, t => {
   var first = true
-  var resolve = require("resolve")
+  var path = require("path")
 
   return compareFixtures(t, "order", {
     path: "fixtures/imports",
-    resolve: (id, base, opts) => {
-      var resolveOpts = {
-        basedir: base,
-        moduleDirectory: [],
-        paths: opts.path,
-        extensions: [ ".css" ],
-      }
-
-      return new Promise(function(res, rej) {
-        var doResolve = () => {
-          resolve(id, resolveOpts, function(err, path) {
-            if (err) {
-              return rej(err)
-            }
-            res(path)
-          })
-        }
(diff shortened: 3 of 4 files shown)


RyanZim:
This has a few problems, do not merge.

RyanZim:
Fixed. @MoOx ?

index.js
```diff
@@ -163,7 +164,7 @@ function parseStyles(
 ) {
   var statements = parseStatements(result, styles)
 
-  return Promise.all(statements.map(function(stmt) {
+  return Promise.resolve(statements).then(promiseEach(function(stmt) {
```
MoOx: what is the difference between promise each and Promise.all? Cause from [the doc](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Promise/all) I can see

>  The array of values maintains the order of the original iterable object, not the order that the promises were resolved in.

RyanZim: @MoOx I should have explained these changes better.

`map` runs the function down to `return resolveImportId()`. This starts a promise chain. `map` then goes on to the next item in `statements`, without waiting for `resolveImportId` to resolve. This causes concurrency. Depending on the order that the files are loaded, `state` is modified incorrectly. If `foo-second.css` loads first, `bar.css` is added to `state.importedFiles`. When `foo-first.css` is loaded, `bar.css` appears to be already loaded, however, it is in the incorrect order in the bundle.

`promiseEach` waits for `resolveImportId` to resolve before going to the next item in `statements`. promise-each is a standalone port of http://bluebirdjs.com/docs/api/promise.each.html.

This is a performance cut; but this is the quickest way to fix this. If someone can figure out a better deduping system that does not require serialized loading, great; I can't ATM.

Hope this is clear.

MoOx: Clear enough! Thanks!

MoOx:
@RyanZim what is your npm nickname?

MoOx:
You know have the full rights for this repo.

To cut a release you will need to:

- be logged on npm from your terminal
- update changelog to add a new section + bump version number in package.json (like explained here https://github.com/MoOx/npmpub#requirements, you don't need to install it, it's already a dev deps of this project)
- commit your changelog add + package.json update with a message being the version number for clarity
- run "npm run release". This should run test with a fresh node_modules (to ensure everything will work correctly from a fresh install), then will release a new version to npm and will cut a github release using notes from the changelog

If you are not confortable doing this, I can do this with you if you want, or I can even do it alone, it will take me a minute.

RyanZim:
@MoOx ryanzim

https://www.npmjs.com/~ryanzim

MoOx:
image

RyanZim:
@MoOx Couple of questions:

- `master` requires the status check __continuous-integration/appveyor/branch__. Isn't that incorrect?
- Should this change be published immediately, or should we add a few other things first? CC: @ai

MoOx:
I think when there are fork, continuous-integration/appveyor/branch is incorrectly not working as expected. No big deal imo.
You can release this asap as a patch. Do not hesitate to cut 3 releases in a day if you feel the need :) As soon as you respect semver (tldr: breaking change: major, new feature: minor, fix: patch).

RyanZim:
> I think when there are fork, continuous-integration/appveyor/branch is incorrectly not working as expected.

Branch only runs on internal PRs.

RyanZim:
@MoOx I would like to do releases in the future; but can you ship this today? I don't have time ATM.

MoOx:
I just released 8.1.3

RyanZim:
Thanks!

Issue #242, opened 2016-11-10
v9.0.0 Discussion


RyanZim:
Attention @ai @MoOx @jonathantneal and anyone else who is interested:

As discussed in https://github.com/postcss/postcss-import/issues/212, we will need to make breaking changes to postcss-import in the near future. That means a new major version.

As @MoOx said elsewhere, it's fine to make 3 releases a day if we have to. I agree with that, but that advice does not apply to major versions IMO. I would prefer not to have to make another major version right after this one.

Does anyone have ideas/recommendations for other breaking changes that we should/could make?

If they can be implemented in a reasonable amount of time, I will hold off on the major release so we can ship all the breaking changes together. If not, I will just ship v9 with no other changes than #212.

Feel free to brainstorm; crazy ideas are OK. :smile_cat:

michael-ciniawsky:
This is the code I started 'fresh' with a few days ago. It's very basic atm, e.g `postcss-load-config`, recursive @imports are missing etc. The reason I post it here is, that imo the current code base of postcss-import is/has become way to complicated. No offense to anyone here :), but for comparison. Let's assume e.g recursion could work with that, maybe a few `root`(path) && `node.params` checks here and there and bam! 😛 `@imports` don't even need to be strictly at the top here, though it is advisable and 'standard'.

```js
const { join, resolve, dirname } = require('path')
const { readFile } = require('fs')

const { assign } = Object

const postcss = require('postcss')

function read (file) {
  return new Promise((resolve, reject) => {
    readFile(file, 'utf8', (err, file) => err ? reject(err) : resolve(file))
  })
}

function process (file, $, { options, plugins }, result) {
  return read(file).then((css) => {
    postcss(plugins)
      .process(css, options)
      .then((_import) => {
        return $.replaceWith(_import.root)
      })
      .catch((err) => console.log(err))
  })
}

function PostCSSImport ({ root = __dirname, options = {}, plugins = [] }) {
  return function (tree, result) {
    root = resolve(root)

    plugins = plugins.length === 0 ? result.processor.plugins : plugins

    options = assign(result.opts, options)

    let imports = []

    tree.walkAtRules('import', (_import) => {
      let _file = join(root, _import.params.replace(/"/g, ''))

      imports.push(process(_file, _import, { options, plugins }))

      result.messages.push({ type: 'dependency', file: _file })
    })
    
    return Promise.all(imports)
  }
}

module.exports = postcss.plugin('postcss-import', PostCSSImport)
```

[...]

PR #245, opened 2016-11-16
Play well with Custom Syntax

RyanZim:
This is still a WIP. Will most likely need rebased when I'm done.

---

_Warning: This is a weird approach to solving this problem._

Perhaps this approach is naive; but it should be pretty well bulletproof. It is based on the (correct :stuck_out_tongue: ) assumption that if a parser can parse the input, it is the correct parser.

If the file is `.sss`, it tries that parser first. For all other files (or if SugarSS parsing fails) it tries the user-passed parser (if available). If that fails, it tries the default parser.

The reason I chose this approach is that we have no reliable way of telling if `result.opts.parser` is a CSS-compliant parser. See https://github.com/postcss/postcss-import/issues/242 for more background.

The main problem with this approach is error handling/reporting. Currently, it gives the user the error from the default parser. In the case of a malformed `.sss` import, this is quite unhelpful. Not sure what is the best route to improve this.

---

This is by no means a polished PR. I may end up totally rewriting the implementation. Anyway, it does provide a good test suite for how things should behave.

CC: @ai @MoOx

### DIFF
--- a/index.js
+++ b/index.js
@@ -4,6 +4,7 @@ var postcss = require("postcss")
 var joinMedia = require("./lib/join-media")
 var resolveId = require("./lib/resolve-id")
 var loadContent = require("./lib/load-content")
+var processContent = require("./lib/process-content")
 var parseStatements = require("./lib/parse-statements")
 var promiseEach = require("promise-each")
 
@@ -314,11 +315,12 @@ function loadImportContent(
       return
     }
 
-    return postcss(options.plugins).process(content, {
-      from: filename,
-      syntax: result.opts.syntax,
-      parser: result.opts.parser,
-    })
+    return processContent(
+      result,
+      content,
+      filename,
+      options
+    )
     .then(function(importedResult) {
       var styles = importedResult.root
       result.messages = result.messages.concat(importedResult.messages)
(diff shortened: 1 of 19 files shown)


ai:
Awesome! You did big work in short time :).

RyanZim:
@ai @MoOx Made a change so that only the SugarSS parser is tried when the extension is `.sss`. That should fix some of the incorrect errors.

You're still stuck with bad errors if you're using a css-compliant parser and make a syntax error. You will get the error message from the default parser. :cry: Thoughts?

ai:
Looks like good. But sorry, I have no `postcss-import` case to test it.

MoOx:
I can't help, I never used anything that raw css.

RyanZim:
I don't use custom parsers either.

Do you think this error problem is big enough to block this PR? We can always change error reporting later.

RyanZim:
@ai @MoOx Not sure if it's worth adding support for `.scss` file extensions or not. What do you think?

ai:
Sure, complicated question. I afraid that user will try to import real SCSS sources (not just with `//` comments).

Let's start without it, and add it later if user will ask for it (at least we have use cases from them).

RyanZim:
@ai That's what I was thinking.

I'll rebase this and then leave it for final review before merging it.

RyanZim:
@ai @MoOx Anything else that needs changed? If not, I will merge this.

RyanZim:
OK, merging.

PR #246, opened 2016-11-16
Add docs for dependency message

RyanZim:
Will merge in a few days if there are no objections.

### DIFF
--- a/README.md
+++ b/README.md
@@ -228,6 +228,20 @@ postcss()
   })
 ```
 
+## `dependency` Message Support
+
+`postcss-import` adds a message to `result.messages` for each `@import`. Messages are in the following format:
+
+```
+{
+  type: 'dependency',
+  file: absoluteFilePath,
+  parent: fileContainingTheImport
+}
+```
+
+This is mainly for use by postcss runners that implement file watching.
+
 ---
 
 ## CONTRIBUTING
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -228,6 +228,20 @@ postcss()
...
+
+```
+{
+  type: 'dependency',
+  file: absoluteFilePath,
+  parent: fileContainingTheImport
+}
+```
+
+This is mainly for use by postcss runners that implement file watching.
```
michael-ciniawsky: Maybe add a explicit note about webpack here, because especially webpack devs are affected by this replacement/enhancement for `addDependencyTo: webpack`

RyanZim: @michael-ciniawsky Just to confirm, the current version of `postcss-loader` listens for this message, correct?

michael-ciniawsky: yep, `postcss-loader` v1.0+ has support for [dependency message](https://github.com/postcss/postcss-loader/blob/master/index.js#L113)

RyanZim: I agree that that fact should be documented; perhaps at `addDependencyTo`'s docs would be a better place.

RyanZim: Done

michael-ciniawsky: Last nitpick 😛 , maybe in both sections

```
This is mainly for use by postcss runners [ ( e.g postcss-loader) ] that implement file watching.[ ( e.g webpack) ] 
```

Issue #192, opened 2016-04-07
Failed to find should be an error, not a warning


call-a3:
Maybe there is a valid use-case for issuing a warning instead of an error that I don't know of. In that case I would suggest adding an option that can disable this "tolerancy" for unfound files.

RyanZim:
@MoOx Thoughts?

MoOx:
Might be a good idea to support that.

RyanZim:
@MoOx I can't see any valid case for warning. File not found is an Error. Will change this for v9 unless you disagree.

MoOx:
From what I am  seeing here, it's supposed to throw https://github.com/postcss/postcss-import/blob/58ae15444661ec4d1c98f4168025a315e42cc5c4/lib/resolve-id.js#L61

RyanZim:
Yeah, but the next `catch` block turns it into a warning: https://github.com/postcss/postcss-import/blob/58ae15444661ec4d1c98f4168025a315e42cc5c4/index.js#L248.

We need to add a `catch` block earlier for this error. Should we `process.exit(1)` for this error?

MoOx:
Han I didn't write this code, it's part of v8 :(
I would go for throwing again, using a setTimeout trick in the catch.

Library code should never ever do process.exit. Only binary should.

RyanZim:
I hadn't thought of `setTimeout`. I'm still not sure if that's a good idea; I can see people complaining about no way to handle the error. `process.exit` isn't a good idea either. Perhaps we should add a special case so that that catch block doesn't turn `file not found` into a warning. That way, we could just let the error trickle up the promise chain. Wouldn't be hard.

MoOx:
In the async world it's not easy. Maybe a option like onError, default to setTimeout that throw? This way by default it crash, and if people are not happy, they can catch the error.

RyanZim:
Not another option!!! Special-casing this error wouldn't be hard; I'll throw together a PR so you can see what I mean.

RyanZim:
Fixed by https://github.com/postcss/postcss-import/pull/247. Will be shipped in v9.

PR #247, opened 2016-11-17
BREAKING: File not found is an Error, not a Warning

RyanZim:
Fixes https://github.com/postcss/postcss-import/issues/192.

@MoOx What do you think of this? Is there some other case that we can test for a readable stack trace, since file not found is now an error?

### DIFF
--- a/index.js
+++ b/index.js
@@ -261,6 +261,7 @@ function resolveImportId(
     }, [])
   })
   .catch(function(err) {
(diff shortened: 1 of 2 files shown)


RyanZim:
@MoOx Any ideas on if/what we should use for testing for a readable stack trace, since I removed the old test?

MoOx:
you can check error.stack, error.message etc

Issue #216, opened 2016-06-06
Resolve function behavior


CookPete:
This is a slightly tricky problem to explain so bear with me.

I am using `resolve` to enable some sort of "theme" system for my stylesheets. I am using a fairly simple `resolve` function:

``` js
postcssImport({
  addDependencyTo: webpack,
  resolve: id => id.replace(/^theme/, PATH_THEME)
})
```

So now I can do:

``` css
@import 'theme/variables.css';
```

and this will, for example, give me the variables from `src/themes/default/variables.css` for my default theme. Great.

The problem happens when _another_ imported file tries to import `variables.css`. For example, I add a `mixins.css` file with:

``` css
@import 'theme/variables.css';
/* Use variables here for mixins */
```

and then I want to use

``` css
@import './mixins.css';
```

in my stylesheets, but it seems like the `resolve` rule is no longer kicking in for the deeper `@import`. Now I get the error:

```
client?843a:47 ./~/css-loader?modules&sourceMap&localIdentName=[hash:base64:3]!./~/postcss-loader!./src/components/Navigation.css
Module build failed: Error: ENOENT: no such file or directory, open '/Users/user/repo/src/theme/variables.css'
    at Error (native)
 @ ./src/components/Navigation.css 4:14-178 13:2-17:4 14:20-184
```

You can see it's trying to open `theme/variables.css` which doesn't exist and isn't being `resolve`ed according to the configuration.

Am I doing something wrong, or is this a bug/limitation of the `resolve` feature?

MoOx:
This may seems like a bug. But to be honest v8 is pretty buggy. See #210.

CookPete:
Ok so I dug around a bit and the [default resolve function](https://github.com/postcss/postcss-import/blob/master/index.js#L14) was no longer kicking in for files that did *not* match my regex. I have changed my code to:
```js
import postcssImport from 'postcss-import'
import postcssImportResolve from 'postcss-import/lib/resolve-id'

// ...

postcssImport({
  addDependencyTo: webpack,
  resolve: (id, base, options) => {
    if (/^theme/.test(id)) {
      return id.replace(/^theme/, PATH_THEME)
    } else {
      return postcssImportResolve(id, base, options)
    }
  }
}),
```

[...]

PR #249, opened 2016-11-19
Change resolve behavior

RyanZim:
Use default resolver when custom resolve does not return an absolute path.

Resolves https://github.com/postcss/postcss-import/issues/216. (No pun intended :smile_cat: )

@ai Does this look good?


RyanZim:
@ai ping?

ai:
LGTM 👍

Issue #195, opened 2016-04-21
Support custom moduleDirectories


taion:
In 7.x, the specified `path`(s) functioned as additional module directories.

I used this to follow the "huge apps" example in using nested `shared/` directories, that were resolved as module directories.

Was there a specific reason for its removal? It's not a feature I've made heavy use of, but it seems like an odd removal. Is the idea that this feature is superseded by the custom `resolve` argument?

MoOx:
> Is the idea that this feature is superseded by the custom resolve argument?

I think so. I have to admit I don't really use this package except for ``node_modules`` and ``web_modules``.

@TrySound handled the 8.x release (huge refactoring). It seems there is some major issues tho like #168 so I may consider simply reverting to 7.x implementation (which is simpler and don't have this kind of bugs) since nobody is currently maintaining this package at the moment (and I found the current codebase more complicated, and API too low level).

taion:
FWIW, this ended up being easy to work around for me, since apparently I was only using the moduleDirectories thing for a single directory.

RyanZim:
> it seems like an odd removal

Agreed.

I would be in favor of adding this back; however, I'm not sure of the best route to proceed. Should custom moduleDirectories replace or be appended to the default `[ 'node_modules', 'web_modules' ]`?

@MoOx @TrySound @ai Thoughts?

MoOx:
I would go for appending them. I guess nobody will want to remove those.

RyanZim:
> I guess nobody will want to remove those.

Yeah, if they do, they can use `resolve`.

neezer:
Voicing my support for this. I'm using the [application structure](https://gist.github.com/ryanflorence/daafb1e3cb8ad740b346) outlined by @ryanflorence, which relies on being able to customize `moduleDirectories`. Unlike @taion, I am actually using this feature pretty heavily. 😉 

In the meantime, I've just forked this repo and [added the directory manually](https://github.com/postcss/postcss-import/compare/master...neezer:master), which unblocks me until any official update.

RyanZim:
@neezer Thanks for chiming in. I want to get v9 shipped today or tomorrow; then this will be PR welcome for the next release.

RyanZim:
@neezer @taion v9 is shipped, PR welcome.

taion:
[...]

PR #256, opened 2016-12-10
Add `addModulesDirectories` config

neezer:
Allows appending to the default list of modules directories. I picked a name that I think sounds good, but happy to change it if y'all have something else you'd prefer. 😄 

Relevant discussion: https://github.com/postcss/postcss-import/issues/195

@RyanZim

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,6 +198,19 @@ postcss: function(webpack) {
 }
 ```
 
+#### `addModulesDirectories`
+
+Type: `Array`  
+Default: `[]`
+
+An array of folder names to add to [Node's resolver](https://nodejs.org/api/modules.html#modules_loading_from_node_modules_folders).
+Values will be appended to the default resolve directories: 
+`["node_modules", "web_modules"]`.
+
+This option is only for adding additional directories to default resolver. If
+you provide your own resolver via the `resolve` configuration option above, then
+this value will be ignored.
+
 #### Example with some options
 
 ```js
--- a/README.md
+++ b/README.md
@@ -203,7 +203,7 @@ postcss: function(webpack) {
 Type: `Array`  
 Default: `[]`
 
-An array of folder names to add to [Node's resolver](https://nodejs.org/api/modules.html#modules_loading_from_node_modules_folders).
+An array of folder names to add to [Node's resolver](https://github.com/substack/node-resolve).
 Values will be appended to the default resolve directories: 
 `["node_modules", "web_modules"]`.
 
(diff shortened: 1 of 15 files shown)


michael-ciniawsky:
Maybe we could refactor `options.resolve` for that , instead of adding it to options directly?

```js
let resolveOpt = {} // defaults

if (typeof options.resolve === 'object') {
   resolveOpt = Object.assign(resolveOpt, options.resolve)
   resolveOpt.moduleDirectories.push('node_modules', 'web_modules')
}

if (typeof options.resolve === 'function') {
  // Use custom resolver
}
```

neezer:
Neat, @michael-ciniawsky! Happy to change the PR to this approach.

@RyanZim @MoOx Any preference?

RyanZim:
@michael-ciniawsky I like that idea, I just think using the `resolve` option is rather implicit behavior.

If we go that direction, we should set `root`, and `path` in the same object as `moduleDirectories`.

For right now, I would be in favor of just doing it the way it is done in this PR. For the next major version, I would consider removing `root`, `path`, and `addModuleDirectories`, in favor of a single `resolveOpts` option; bikeshed about that later.

README.md
```diff
@@ -198,6 +198,19 @@ postcss: function(webpack) {
 }
 ```
 
+#### `addModulesDirectories`
+
+Type: `Array`  
+Default: `[]`
+
+An array of folder names to add to [Node's resolver](https://nodejs.org/api/modules.html#modules_loading_from_node_modules_folders).
```
RyanZim: Probably should link to https://github.com/substack/node-resolve instead, since that is the module that is actually used.

neezer: @RyanZim Updated!

RyanZim:
@ai Thoughts? @MoOx Any objections?

RyanZim:
@neezer Thanks!

renatorib:
It's not published yet?

RyanZim:
@renatorib It's not. I should publish it. I was planning on adding a few other things to the next release, but I haven't got around to it, so I guess I'll just publish this. I'll comment here as soon as I get it done.

RyanZim:
@renatorib @neezer Released: https://github.com/postcss/postcss-import/releases/tag/9.1.0. Sorry for the long wait.

neezer:
@RyanZim Great! So should #195 be closed now?

RyanZim:
@neezer Yep, thanks.

Issue #254, opened 2016-12-09
Add option to blacklist paths


IonicaBizau:
Let's say I want to blacklist a specific url, which is added in the result, inlined, currently.

How can I do that?

RyanZim:
@IonicaBizau I'm don't think that is possible currently. Wouldn't be opposed to adding that, would just need to add some conditional logic around https://github.com/postcss/postcss-import/blob/master/index.js#L182.

Won't have time to work on this in the near future, PR welcome.

We should probably accept a string or regex; or an array of strings or regexes to match against.

IonicaBizau:
@RyanZim I did try to hack into the code and specifically I already tried to change that line, but it just didn't work... It was still replacing the `@import`s with the content.

I'd be happy to add it, but will need some guidance where that should be added... 😁 

Thanks!

RyanZim:
Adding another `if` statement with an empty `return` if the path is blacklisted should do it, not sure what your issue was. Right now I'm working 10 & 12-hour days at my day job, so no time to look at this now.

IonicaBizau:
It turns out that this tool is smart enough to know when it should inline the things and when not. Therefore a blacklist is not critical for me now.

Still, it'll be nice to have in the future.

RyanZim:
Fixed in #327

PR #259, opened 2017-01-11
Add `beforeImport` option

finian:
for whitelisting or blacklisting paths  #254

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,6 +133,30 @@ Default: `undefined`
 
 An array of plugins to be applied on each imported files.
 
+#### `beforeImport`
+
+Type: `Function`
+Default: `null`
+
+Function called before the import process. Take one argument `path` and should
+return a path or a falsy value. If a falsy value is returned, the import statement
+will not be processed.
+You can use this for whitelisting or blacklisting paths.
+
+e.g. Only a path prefixed with `#` symbol will be processed:
+
+```
+{
+  beforeImport: function (path) {
+    if (/^#/.test(path)) {
+      return path.substring(1)
+    }
+
+    return false
+  }
+}
+```
+
 #### `onImport`
 
 Type: `Function`  
--- a/index.js
+++ b/index.js
@@ -178,9 +178,15 @@ function parseStyles(
     stmt.media = joinMedia(media, stmt.media || [])
 
     // skip protocol base uri (protocol://url) or protocol-relative
-    if (stmt.type !== "import" || /^(?:[a-z]+:)?\/\//i.test(stmt.uri)) {
+    if (stmt.type !== "import" || /^(?:[a-z]+:)?\/\//i.test(stmt.uri) ||
+        options.beforeImport && !options.beforeImport(stmt.uri)) {
       return
     }
+
+    if (options.beforeImport) {
+      stmt.uri = options.beforeImport(stmt.uri)
+    }
+
     return resolveImportId(
       result,
       stmt,


RyanZim:
Not a bad idea; @ai @MoOx thoughts on this?

MoOx:
Looks overkill to me. If some url should be skipped, it should be built-in.

ai:
Maybe some PostCSS plugin before `postcss-import` is better solution?

RyanZim:
@ai What would the plugin do?

ai:
Find `@import` with blacklisted paths and remove them?

RyanZim:
@ai I'm not sure what the original intent was; I assumed @IonicaBizau wanted to allow the browser to load some paths (instead of inlining them dev-side). Perhaps I'm wrong; @IonicaBizau?

IonicaBizau:
@RyanZim You're right. My idea was that the blacklisted paths would remain as import lines, letting the browser to load them.

finian:
Hi all, thanks for reviewing. My use case: I use webpack to load css (it processes @import things) and use postcss-cssnext to write css. In a css file, I want to import global variables, so I use postcss-cssnext to inline them, but I don't want other @import statements to be inlined (so they can be processed by webpack).

```css
/* variables.css */
:root {
  --mainTextFontSize: 14px;
}


/* module.css */
@import './variables.css'; /* <-- inline this */
@import './module2.css'; /* <-- leave this alone, don't inline it */

.body {
  font-size: var(--mainTextFontSize);
}
```

michael-ciniawsky:
@finian Why should `css-loader` handle the @import instead 😛  ?  
@RyanZim Is `@import http(s)://...` for loading in the browser, not ignored by default atm 👀 ?

RyanZim:
@ai In light of this info, what are your thoughts?

I'm not sure if accepting a function is the right way to go, accepting an array of paths or perhaps a regex would work in most cases.

---

Partially related: I've been thinking for some time about requiring use-cases like this to be implemented as a custom `resolve` function. However, setting up a basic resolver is a bit of work. What if we moved `lib/resolve-id.js` into a separate module so that it could be easily extended?

We could also encourage the creation of "resolver modules," that you could use like: `opts.resolve: require('some-resolver')`; and have a page in the wiki with a list of custom resolver modules that would be useful to other users.

Just an idea.

RyanZim:
@michael-ciniawsky Didn't see your comment while I was writing.

> Why should css-loader handle the @import instead :stuck_out_tongue: ?

Good question!

> Is `@import http(s)://...` for loading in the browser, not ignored by default atm :eyes: ?

`@import "http(s)://..."` & `@import "//url.com"` are ignored currently. If users want to import remote stylesheets, they can use https://github.com/unlight/postcss-import-url.

michael-ciniawsky:
> Good question!

Yep, @webpack we want to try redesign/improve the CSS pipeline where possible, curious of the usecase/why here 😛 

> @import "http(s)://..." & @import "//url.com" are ignored currently. If users want to import remote stylesheets, they can use https://github.com/unlight/postcss-import-url.

Yep, I though so, just uncertain for a moment if I saw it @postcss-partial-import or here a while ago

ai:
@RyanZim not this callback looks better for me :D

RyanZim:
@ai So would you merge this?

I'm still not sure if I like the naming.

IonicaBizau:
@RyanZim I recently opened https://github.com/postcss/postcss/issues/951. It's about resources via paths.

It may be related to this PR. Linking it here just in case.

RyanZim:
@ai Thoughts on this?

ai:
LGTM

michael-ciniawsky:
`options.filter` ? Still not convinced that this is actually needed and is more usecase specific as it seems (`webpack`), while it could be avoided otherwise, but not my call to make 😛

RyanZim:
OK, I guess we can accept this if it's renamed to `options.filter`

RyanZim:
Closed by  #327

PR #273, opened 2017-03-24
Added definitions for typescript developers.

OctoD:


RyanZim:
I am hesitant to support this, since I don't use TypeScript, and can't update the typings.

@ai Opinion on this?

ai:
I accept such PR because outdated definitions is not a big problem

ai:
But in other hand I don't like that we increase node_modules size for all even non-TS users. So it is up to you 😆.

RyanZim:
Sorry, I think I'm going to pass on this. I've accepted TypeScript definitions for another project and it didn't turn out so well. Perhaps submit this to https://github.com/DefinitelyTyped/DefinitelyTyped?

Issue #278, opened 2017-04-25
Recursive imports fail silently, transpile empty CSS


jedmao:
With the following configuration options:

```js
{
  path: ['blocks']
}
```
And a folder structure like this:
```
styles
├── blocks
│   ├── foo.css
│   ...
├── foo.css
...
```
If I have a file at `styles/foo.css` that looks like this:
```css
@import 'foo';
```
It transpiles an empty CSS file w/o errors.

As a work-around, I can do this:
```css
@import 'blocks/foo';
```
But I'm kinda' thinking this should throw an error!

RyanZim:
Ugh, I never use `path` myself. What happens if you `path.resolve` `'blocks'`?

jedmao:
The path is not the problem. It resolves just fine. The problem is the recursive import, since they are both named foo.

RyanZim:
Ah, OK; I'll look into this as soon as I get a chance.

RyanZim:
Hey there, I'm terribly busy. If you can, a reduced test case repo on GH would sure help me get around to this faster.

jedmao:
@RyanZim, see #280

RyanZim:
See https://github.com/postcss/postcss-import/pull/280#issuecomment-298134082

jedmao:
Excellent! I just need to reorder my paths!

PR #280, opened 2017-04-27
Break tests w/ recursion example

jedmao:
Related to #278 

@RyanZim here's a breaking test for you!

### DIFF
--- a/test/fixtures/imports/modules/resolve-path-modules.css
+++ b/test/fixtures/imports/modules/resolve-path-modules.css
@@ -1,0 +1,1 @@
+resolve-path-modules {}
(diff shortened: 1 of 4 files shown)


RyanZim:
Thanks!

RyanZim:
@jedmao Finally got a chance to try locally.

postcss-import is behaving as intended. Your test fails because `empty {}` is not included in the output.

This is because `@import "empty";` resolves to `test/fixtures/imports/empty.css`, which is an empty file. You wrote this intending to import `test/fixtures/imports/modules/empty/index.css`.

If you change the order of your `path` array, the test passes:
```js
test("should resolve local modules", t => {
  return compareFixtures(t, "resolve-path-modules", {
    path: [
      "fixtures/imports/modules",
      "fixtures/imports",
    ],
  })
})
```
Will pass

Going to close this as this is intended behavior. Thanks for trying, though.

PR #281, opened 2017-05-06
Remove promise-each dependency

### LINKED ISSUE #275: TypeError: Promise.resolve is not a function
If i don't have `es6-promise` package i get error:
```shell
TypeError: Promise.resolve is not a function
        at /path/to/node_modules/promise-each/index.js:16:16
```

I can do PR if your accept drop `promise-each`.

jonathantneal:
- Resolves #275
- Removes any-promise subdependency
- Removes runtime assert from promise-each

### DIFF
--- a/index.js
+++ b/index.js
@@ -6,7 +6,6 @@ var resolveId = require("./lib/resolve-id")
 var loadContent = require("./lib/load-content")
 var processContent = require("./lib/process-content")
 var parseStatements = require("./lib/parse-statements")
-var promiseEach = require("promise-each")
 
 function AtImport(options) {
   options = assign({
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -174,20 +173,26 @@ function parseStyles(
 ) {
   var statements = parseStatements(result, styles)
 
-  return Promise.resolve(statements).then(promiseEach(function(stmt) {
-    stmt.media = joinMedia(media, stmt.media || [])
+  return Promise.resolve(statements).then(function(stmts) {
+    return stmts.reduce(function(promise, stmt) {
+      return promise.then(function() {
```
RyanZim: LGTM, but perhaps we should have a comment here to explain that this cleverness works on the array in series.

jonathantneal: I’ll add a comment. I’m actually glad to see the `reduce` and `then` terminology used here instead of `each`, because `each` did not make me aware that this was processing the statements in sequence.

RyanZim: @jonathantneal ping?

jonathantneal: Done.

RyanZim:
Thanks!

PR #282, opened 2017-05-10
Bump PostCSS version to 6.0.0

thebuilder:


RyanZim:
We should do a hard update to postcss v6, drop 0.12 support and release a new major version.

RyanZim:
Closing in favor of https://github.com/postcss/postcss-import/pull/283

PR #288, opened 2017-06-01
improve usage example

marzelin:
I found usage example a bit confusing (to make it work `foo.css` and `bar.css` have to be in `./css/css` folder O.o)

I added project structure to make things clearer and made import paths resemble real-world use cases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,11 +44,23 @@ $ npm install postcss-import
 
 ## Usage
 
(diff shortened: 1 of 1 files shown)


RyanZim:
I need to look at this and do a bit of thinking here.

RyanZim:
Fixed in 99fd0f8

PR #292, opened 2017-06-10
Remove transform option from README.md

ZeroCho:
As far as I know, transform option is removed, isn't it? Or can I still use it only in this case?


RyanZim:
@ZeroCho Good catch!

PR #299, opened 2017-08-19
Adding filter method

hirokiosame:
This is a continuation of https://github.com/postcss/postcss-import/pull/259 since it seems to be abandoned. I have updated the name to `filter`.

If the logic/implementation looks good, I will go ahead and add documentation.

Example postcss config file:
```
module.exports = {
	plugins: [
		require('postcss-import')({
			filter(url) {
				if (url.match(/font\.css$/)) {
					return false;
				}
			},
		}),
	],
};
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -140,6 +140,17 @@ function parseStyles(result, styles, options, state, media) {
             return
           }
 
+          if (options.filter) {
+            let filtered = options.filter(stmt.uri);
+
+            if (filtered === false) {
+              return;
+            }
+            if (typeof filtered === "string"){
+              stmt.uri = filtered;
+            }
+          }
+
           return resolveImportId(result, stmt, options, state)
         })
       }, Promise.resolve())
--- a/index.js
+++ b/index.js
@@ -146,9 +146,6 @@ function parseStyles(result, styles, options, state, media) {
             if (filtered === false) {
               return;
             }
-            if (typeof filtered === "string"){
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -140,6 +140,14 @@ function parseStyles(result, styles, options, state, media) {
             return
           }
 
+          if (options.filter) {
+            let filtered = options.filter(stmt.uri);
+
+            if (filtered === false) {
+              return;
+            }
```
RyanZim: Can be shortened to `if (!options.filter(stmt.uri)) return`

RyanZim:
Needs tests, not 100% sure that this will work correctly, will have to do a bit of testing myself.

index.js
hirokiosame: I do a strict `false` check so blocking is deliberate. When the filter has no return-statement and the return value is `undefined`, the filter will have no effect.

I can shorten it to `if (options.filter(stmt.uri) === false)` if you want but I typically opt for the verbose option as readability -> maintainability especially in collaborative projects.

hirokiosame:
@RyanZim What cases are you suspicious of/interested in handling? I'll try to get around to adding tests this weekend.

RyanZim:
@hirokiosame Ping on the tests?

0xcaff:
It looks like this one is dead too. Can I pick this PR up?

RyanZim:
Closed by #327

PR #300, opened 2017-08-23
Fixed: do not preserve symlinks as per Node's default behaviour

cyberthom:
There's a bug in the current version of PostCSS import plugin that occurs when using symlinks and causes invalid dependencies to be loaded in when using different versions of the same dependency within the same app. 

The bug occurs as the default behaviour of the `resolve` module (used to resolve ids to filenames) is different to the default behaviour of Node.js's native `require.resolve` method. 

I know this is probably a bit of an edge case as most people don't use symlinks but it causes huge issues in monorepos (e.g. lerna / yarn workspaces).

The fix is to disable the following option in the `resolve` module:
* `opts.preserveSymlinks` - if true, doesn't resolve basedir to real path before resolving. This is the way Node resolves dependencies when executed with the `--preserve-symlinks` flag. Note: this property is currently true by default but it will be changed to false in the next major version because Node's resolution algorithm does not preserve symlinks by default.
(https://www.npmjs.com/package/resolve)

### DIFF
--- a/lib/resolve-id.js
+++ b/lib/resolve-id.js
@@ -24,6 +24,7 @@ module.exports = function(id, base, options) {
       else if (!pkg.main || !/\.css$/.test(pkg.main)) pkg.main = "index.css"
       return pkg
     },
+    preserveSymlinks: false,
   }
 
   return resolveModule(`./${id}`, resolveOpts)


RyanZim:
This is a crazy edge-case, but I agree it's worth fixing. However, technically, this is a breaking change (semver-major). I'd like to release a few other things in a major release other than this.

cyberthom:
Yeh I agree - It should probably be a major version bump just to make sure no one's relying on the previous behaviour.

Why being precious about version numbers? :)

RyanZim:
There's just other semver-major stuff I've been thinking of doing. I don't mind releasing new versions, but major versions require manual effort for everyone to upgrade, so I don't like to do them too crazily often.

cyberthom:
How about exposing it as an option and leaving the default as is? Happy to modify the PR.

cyberthom:
I've updated the pull request.

Hopefully this will work for you, keeping current behaviour as is but adding a fix for those who want Node's default dependency resolution behaviour.

RyanZim:
Sorry for being slow about getting back. Could you revert this PR back to it's original state? I'm not a fan of having too many options. We'll just release a major for this.

cyberthom:
No worries - all done!

RyanZim:
Thanks, will merge and release as soon as I can.

RyanZim:
Man, just noticed I left this hanging. Will set a reminder and try to get on this ASAP. Sorry about this.

RyanZim:
Landed in 80cadec2087c0fcdb310c53de00e38e1af4305ca.

RyanZim:
Publishing in v11 now

cyberthom:
Awesome - Thanks!!

PR #306, opened 2017-10-17
Adding filename to resolver arguments

BaliBalo:
It would be nice to be able to know from which file the import is made in the custom resolver. Currently it only sends the directory name of this file.
I would suggest sending the filename as another argument.

### DIFF
--- a/index.js
+++ b/index.js
@@ -184,7 +184,7 @@ function resolveImportId(result, stmt, options, state) {
       // Ensure that each path is absolute:
       return Promise.all(
         paths.map(file => {
-          return !path.isAbsolute(file) ? resolveId(file, base, options) : file
+          return !path.isAbsolute(file) ? resolveId(file, base, options, path.basename(atRule.source.input.file)) : file
         })
       )
     })
--- a/index.js
+++ b/index.js
@@ -178,13 +178,13 @@ function resolveImportId(result, stmt, options, state) {
     ? path.dirname(atRule.source.input.file)
     : options.root
 
-  return Promise.resolve(options.resolve(stmt.uri, base, options))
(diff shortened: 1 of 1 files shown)


RyanZim:
What's the real-world use-case for this?

BaliBalo:
As for `baseDir`, being able to import a different file based on the request origin file. In my case:

We have an "override" structure with inheritance, and I want the resolving to take the file from any level under the current level, but only if I'm requiring the same file as I am currently in.
Consider this example structure:
```
base/
├─ a.css
├─ b.css
└─ level_1/
   ├─ a.css
   └─ level_2/
      ├─ a.css
      └─ b.css
```

If I require a.css from any js file in the project it should resolve to the deepest one (here `level_2/a.css`).
But I want to be able to "extend" from the other overrides, that means that when I do `@import "a.css"` from `level_2/a.css` I need to get the file from a level below, here `level_1/a.css` (just getting the deepsest one would get itself, which doesn't work, and I don't want the one in `base`, just 1 level below if available).
But if I use `@import "b.css"` from a.css, I want to get the deepest b possible, even if it is the same level, not one from a level below.
So the folders I am looking in depends on the file (if it is the same as the one it was requested from or not).

michael-ciniawsky:
Just keep `.js` && `.css` files coupled

```
|– src
| |– a
| | |– a.css
| | |– a.js // import styles from './a.css'
| |– b
| | |– styles
| | | |– b1.css
| | |– b.css // @import './styles/b1.css' (postcss-import)
| | |– b.js // import styles from './b'
``` 

And use `postcss-import` `@import` only within `module` scope (`folder`), to avoid duplicates

> ⚠️  Keeping all CSS 'co-located'  in a webpack (?) setup is 'bad practice' and has caveats

RyanZim:
Gonna close this as per @michael-ciniawsky

PR #324, opened 2017-12-11
Change from filename to checksum duplicates

samuelthomps0n:
When saving files to the list of includedFiles it currently uses the
full file path which doesn't work well in monorepos. I've changed to
store an MD5 hash of the file so that different version files can be
used, but they can live in different directories.

An example of where this is causing problems:

We have a monorepo that has a shared 'core' for the CSS. We import this in to every package but as its a dependency the file path can often be <package-name>/node_modules/core/src/index.css - currently with the file path check it sees every one as a different file.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,6 +1,7 @@
 "use strict"
 // builtin tooling
 const path = require("path")
+const md5File = require("md5-file")
 
 // external tooling
 const postcss = require("postcss")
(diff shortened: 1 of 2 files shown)


RyanZim:
@samuelthomps0n This is a problem, but I'm not sure if this is the right solution.

---

> We import this in to every package

Are you copying the files or symlinking them?

samuelthomps0n:
We're using Lerna so theyre being symlinked

RyanZim:
OK, so why don't we just stick in some logic to resolve the symlink to the absolute file path and use that for the dupe logic? (i.e. `fs.realPath()`)

samuelthomps0n:
I'll take a look at doing that in the morning :)

samuelthomps0n:
Had a look at this this morning - turns out they're not symlinks. When we pull the packages in to the main project it will install each dependency in to each package, like below:

- package-a
- package-b
  - /node_modules/package-a

Our main CSS file in the site imports both package A and package B, so once we've got package A we don't need it again, but as package-b is @import'ing it gets loaded in again. Thats why I looked at checking with a file hash as if the file is identical there's no need to load it again, irrespective of the file path.

RyanZim:
Can you make a minimal test case repo that demonstrates your setup and the duplication you're having? I'm not 100% sure I understand it, and I'd like to fool with this locally.

Just so you know, the reason I'm balking about this hashing solution is that it's going to make a significant performance hit (postcss-import is already famous for being one of the slowest postcss plugins). If there's any other way we can solve this, I want to go with that solution.

samuelthomps0n:
Interestingly I've just created a demo repo with exactly the same node_modules set up as our live project and skipDuplicates is working exactly as expected! Very frustrating!

Im going to close this Pull Request for now and see if I can pinpoint exactly what is causing our issues as clearly it has changed since last time I looked at it. It may well be one of our other plugins, or the fact that everything is run though Webpack.

Thanks for your help though :)

Issue #179, opened 2016-03-02
Allow import to work within :global


lucasmotta:
Hey there,

Recently me and my co-workers made the move to use [css-modules](https://github.com/css-modules/css-modules) and we are having our up and downs with it.

One of the things we came across is importing vendors, so we decided to use `postcss-import`, which works wonder.  
But our problem is that `css-loader` hashes the class name of everything (we kind of want that). And since `postcss-import` happens before the `css-loader`, there's no really way to exclude the vendor from the `css-loader` process.

We found a solution, which is to have a `globals.css` file and create two configs on our webpack config, but don't apply the `css-loader` modules only on the `globals.css`. It works, but it's not ideal.
We could as well request the creators of those vendors to wrap their css on  a `:global {}` ns, but that brings even more complications.

So our suggestion was to allow `@import` inside the `:global {}` ns, something like this:

```scss
// app.css
:global {
  @import "vendor-1";
  @import "vendor-2";
}
```
Which would compile to:
```
:global {
  .vendor1 {
    content: "hey I'm vendor 1";
  }
  .vendor2 {
    content: "and I'm vendor 2!";
  }
}
```

I don't see this creating any compatibility problems with current users, but that would help a lot of people who are adventuring on the `css-modules` world.

If this is something you guys are happy with it, I could help implement it too :)

MicheleBertoli:
+1

MoOx:
Am I right assuming you use webpack guys?

lucasmotta:
@MoOx yes, that's correct! Currently we have two configs for the webpack loader for the time being:
```js
[
  {
    test: /\.css$/,
    loader: 'style!css?modules&importLoaders=1!postcss',
    exclude: /global/,
  },
  {
    test: /\.css$/,
    loader: 'style!css!postcss',
    include: /global/,
  }
]
```

lucasmotta:
We even thought about creating another plugin that introduces another keyword, something like `@global` that would wrap the content of the vendor on `:global {}`:

```css
@global "vendor-1";
@global "vendor-2";
```

But we would rather try to implement this on `postcss-import` instead of introducing something new.

TrySound:
@lucasmotta Try this https://github.com/jonathantneal/postcss-partial-import

TrySound:
:global is not part of spec (Am I wrong?). In this case we can't introduce it.

MoOx:
I have this "issue" with webpack so I use 2 config section, and as you said it works.

[...]

PR #327, opened 2018-01-02
Added Filter Parameter

### LINKED ISSUE #279: Nested imports have wrong path
```
/component1/index.sss
/component2/index.sss
```

```
// component1
@import '@scope/module/file.sss'
```

```
// component2
@import '../component1/index.sss'
```

The `component2` will throw an error:

```
Module build failed: Error: ENOENT: no such file or directory, open '/Users/FezVrasta/app/component2/@scope/components1/index.sss'
```

### LINKED ISSUE #297: Filtering imports
I'd like to be able to filter (with whitelist/blacklist/custom-function) the files that are imported by postcss-import so that certain files can be resolved by the webpack css-loader instead (eg. for unique modules). I tried passing in a `load` function but I may be doing it incorrectly:
```
require('postcss-import')({
	load: (function () {

		const loadContent = require('postcss-import/lib/load-content');

		return function (filename, options) {
			if (filename.match(/module-for-css-loader/)) {
				return `@import "~module-for-css-loader";`;
			}

			return loadContent(filename, options);
		};
	})(),
}),
```

0xcaff:
This allows for conditionally processing imports based on their URLs.

Closes #297, #254, #299, #259 

Docs and tests have been added.

This PR will help solve the following issues:

https://github.com/postcss/postcss-import/issues/179
https://github.com/css-modules/css-modules/pull/65


README.md
```diff
@@ -110,6 +110,14 @@ Checkout the [tests](test) for more examples.
 
 ### Options
 
+### `filter`
+Type: `string => boolean`  
+Default: `() => true`
+
+Only transforms imports for which the test function returns true. Imports for
+which the test function returns false will be left as is. The argument to the
+function is the url from the import (`@import (url)`).
```
RyanZim: Revise to:

> Type: `Function` 
> Default: `() => true`
> 
> Only transform imports for which the test function returns `true`. Imports for
which the test function returns `false` will be left as is. Function gets the path to import as an argument and should return a boolean.

0xcaff:
Why would this be a concern? Import statements must precede all other css statements except `@charset`? I don't see any other way this could work. I think this behavior is expected and should just be documented.

RyanZim:
CSS is order-sensitive; but since imports must be at the top, we don't really have alternatives here.

> should just be documented

Probably right here.

0xcaff:
I've added the note to the readme about import order. Anything else?

RyanZim:
Sorry about the wait here, not sure how this slipped through the cracks. I'll try to merge and release soon.

PR #337, opened 2018-01-23
Added RxJS Style NPM packages and tests

oleersoy:
Added tests that show the ability of postcss-import to consume RxJS style subpackages.  In other words instead of consuming the entire kitchen sink via the `style` or `main` attribute, only consume a subpackage of the npm module.


RyanZim:
Thanks!

PR #380, opened 2018-10-19
Set plugin property to dependency messages which is required according to the API doc

### LINKED ISSUE #379: postcss-import doesn't set plugin property to its dependency messages
According to the postcss API document, [`plugin` is required property in Result#messages](http://api.postcss.org/Result.html#messages), but [`postcss-import` doesn't set it when it pushes dependency information to the messages](https://github.com/postcss/postcss-import/blob/f98dd1a5a6a987c8cf197db46644765a4d43d73a/index.js#L198-L205).

nodaguti:
This PR fixes #379. The lack of `plugin` property leads to a message mishandling by other plugins and may cause an issue described as the second problem in https://github.com/nodaguti/postcss-calc-warning.

### DIFF
--- a/index.js
+++ b/index.js
@@ -199,6 +199,7 @@ function resolveImportId(result, stmt, options, state) {
       resolved.forEach(file => {
         result.messages.push({
           type: "dependency",
+          plugin: "postcss-import",
           file: file,
           parent: sourceFile,
         })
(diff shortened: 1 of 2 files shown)


nodaguti:
@RyanZim Thank you too!

PR #389, opened 2019-01-23
Rewrite `composes` rules inside @imported stylesheets

vitalif:
Hi. There is a problem with postcss-import that it doesn't rewrite relative paths in `composes: x from '../file'` when inlining `@import`ed files, which breaks this kind of composition in css modules. I think the best place to fix it is postcss-import itself; also I couldn't find a way to fix it by configuration of any other loaders or plugins.
I've prepared two versions of this fix - for master and for 11.1.0. Could you please merge it?


vitalif:
Oops, sorry, it's all github picking the bad branch. ` rewrite-composes-11.x ` is of course for 11.1. Pick this commit for master: https://github.com/vitalif/postcss-import/commit/6fce668bb9b3e7ca4313993b5afdb3299d42d0ef

RyanZim:
We do not formally support css-modules, so this is out of scope.

PR #407, opened 2019-12-27
caseSensitive option to trigger failures due to case-sensitive file system paths

MarcusNTuttle:
While using postcss-import, I found that imports and builds were failing once we uploaded them to the server, but not locally. This was because I was using a noncase-sensitive file system (HFS+) while the server was using a case-sensitive file system (NTFS). This option should allow local builds to fail due to case-sensitive paths regardless if the dev's file system is case-sensitive or not.

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,6 +189,16 @@ This option is only for adding additional directories to default resolver. If
 you provide your own resolver via the `resolve` configuration option above, then
 this value will be ignored.
 
+#### `caseSensitive`
+
+Type: `Boolean`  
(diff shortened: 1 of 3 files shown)


RyanZim:
Sorry for the slow response here; I'm not sure how I missed this. Dealing with case-sensitivity is the responsibility of the developer; it's not standard practice for tooling to do anything about this. Doing it here is of little use, given that postcss itself defers case handling to the underlying filesystem. Sorry, but will not be accepting this.

PR #410, opened 2020-03-11
Add TypeScript typings

remcohaszing:
This makes the package usable in TypeScript projects.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,64 @@
+import { AcceptedPlugin, Transformer } from "postcss";
+
+interface AtImportOptions {
+  /**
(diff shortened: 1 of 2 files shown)


RyanZim:
The problem here is, I don't know typescript myself, so I have no way to keep them updated.

remcohaszing:
If you prefer not to include them, I’ll make a pull request to DefinitelyTyped instead. This way people can install the types as a separate package. That’s the way to go if package maintainers don’t want to maintain the typings.

Just let me know :smile:

RyanZim:
Yeah, I'd prefer DefinitelyTyped. Thanks!

PR #441, opened 2020-11-12
Added github action to convert to Deno

oscarotero:
This PR adds a github action that is runned after creating new tags, so it will update automatically the code in the https://github.com/postcss/postcss-deno-import repository (the Deno version of the plugin) and create the same tag there.

The only requirement is a secret variable with a [PAT](https://docs.github.com/en/free-pro-team@latest/github/authenticating-to-github/creating-a-personal-access-token) that have writting access to postcss-deno-import repository in order to commit the changes and create the new tag. So, after merging this PR, you have to:

- Go to Settings > Secrets of this repository
- Click in "New Secret"
- The secret name is `PAT`, and the value the access token

That's all.

### DIFF
--- a/.github/workflows/deno.yml
+++ b/.github/workflows/deno.yml
@@ -3,6 +3,7 @@ name: Deno
 on:
   push:
     tags:
+      - "*"
 
 jobs:
   build:
--- a/.github/workflows/deno.yml
+++ b/.github/workflows/deno.yml
@@ -3,7 +3,6 @@ name: Deno
 on:
   push:
     tags:
-      - "*"
 
 jobs:
   build:
@@ -11,7 +10,7 @@ jobs:
 
     steps:
       - name: Setup Deno environment
-        uses: denolib/setup-deno@v2.2.0
+        uses: denolib/setup-deno@v2.3.0
         with:
           deno-version: v1.x
 
@@ -32,9 +31,6 @@ jobs:
           rm -rf deno
           deno run --unstable --allow-write --allow-read to_deno.js
           deno fmt deno
-          cp postcss/README.md deno/
-          cp postcss/CHANGELOG.md deno/
-          cp postcss/LICENSE deno/
 
           deno test --unstable --allow-read deno/test/*
 
--- a/.github/workflows/deno.yml
+++ b/.github/workflows/deno.yml
@@ -1,0 +1,53 @@
+name: Deno
+
+on:
+  push:
+    tags:
+      - "13.*"
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Setup Deno environment
+        uses: denolib/setup-deno@v2.2.0
+        with:
+          deno-version: v1.x
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -219,6 +219,19 @@ postcss()
...
+This plugin is available for [Deno](https://deno.land/):
+
+```js
+import postcss from "https://deno.land/x/postcss/mod.js";
+import postcssImport from "https://deno.land/x/postcss_import/mod.js";
+
+const result = await postcss([postcssImport]).process(css);
+```
+
+Note that the Deno version has not support for module resolution like Node (packages installed with npm or folders with a `package.json` file).
```
RyanZim, line 233:
```suggestion
Note that the Deno version does not support Node-like module resolution (packages installed with npm or folders with a `package.json` file).
```
Clarified sentence. 

Just so I understand, that means that in deno, you can only use imports like `@import "./path/to/file.css"`?

RyanZim:
In regards to the personal access token stuff, I'm not comfortable putting a PAT for my account into the secrets here, especially when I haven't fully vetted the toolchain used in the action. @ai how does postcss repo do this? Should we have a bot GH account for this?

.github/workflows/deno.yml
```diff
@@ -0,0 +1,53 @@
+name: Deno
+
+on:
+  push:
+    tags:
+      - "13.*"
```
RyanZim, line 6: Will this line need updated when we bump major versions here?

README.md
oscarotero: Node version uses the `resolve` package as a dependency for all resolutions. I didn't found any deno package that does that, so I've replicated the functionality here: https://github.com/postcss/postcss-deno-import/blob/master/deps.js#L11

For now, it have the following support:

- `@import "./styles.css"` - Obviously
- `@import "styles.css"` - Search the file as relative path (like`./styles.css`) and in all `options.paths`
- `@import "styles"` - Detect `./styles.css` or `./styles/index.css` in the relative path and all `options.paths`.

### What is missing?

It does not load `package.json` files to detect the file in the `style` or `main` fields. This feature could be added but due this is a Deno package, I don't think it makes sense load Node packages here, so I decided don't complicate the code.

.github/workflows/deno.yml
oscarotero: Yep. The `push` event is fired after creating new tags and new branches. So I want to make sure that it's fired only on create a new version tag. To include all tags, we can try with `"*"`.

RyanZim: Yeah, I never tag things that aren't releases, so `*` will work.
```suggestion
      - "*"
```

romainmenke:
Is this still needed?

https://deno.land/manual@v1.35.2/node

Deno now has improved support for `npm` packages.

oscarotero:
It's not needed. I'm using the `npm` version on Deno without problems (https://github.com/lumeland/lume/blob/master/deps/postcss.ts)

I'm closing this.

romainmenke:
Thank you for the fast response 🙇

PR #447, opened 2020-12-12
Fix charset handling

RyanZim:
OK, this is a _slightly_ naive implementation, but I think it's pretty good. If there's a `@charset` anywhere in the CSS, move it to the top of the output and apply it to everything. If there are multiple `@charset`s, compare them. If they're all the same, just use one. If they're different, there's no way we can put all of these in one file, so error out. If you legitimately need to import files with different charsets, you can use the `plugins` or `load` option to add logic to convert all your files to a single encoding that matches the entry point.

Also added checks to ensure that `@charset` comes first, warn the user if this isn't the case.

Technically a breaking change due to adding the error and warning cases.

Attn @SLKnutson

### DIFF
--- a/index.js
+++ b/index.js
@@ -107,10 +107,7 @@ function AtImport(options) {
 
         // Strip additional statements.
         bundle.forEach(stmt => {
-          if (stmt.type === "import") {
-            stmt.node.parent = undefined
-            styles.append(stmt.node)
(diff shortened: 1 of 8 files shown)


SLKnutson:
This looks right to me 👍

RyanZim:
Published in v14.0.0 :tada:

PR #449, opened 2021-02-14
Fix conflict with tailwindcss when using @tailwind before @import

thunderwp:
Hello there
This PR will fix the conflict between postcss-import and tailwindcss when we try to use import after tailwind's @tailwind base or @tailwind components
well obviously we can put @import before these two statements but in the result our custom CSS will be added "before" normalize styles and that is not pretty or standard.

After this PR the above syntax is valid

<pre>
@tailwind base;
@tailwind components;

@import "./buttons.css";
</pre>

Sorry for my bad English.

### DIFF
--- a/lib/parse-statements.js
+++ b/lib/parse-statements.js
@@ -85,7 +85,7 @@ function parseImport(result, atRule) {
       if (
         prev.type !== "comment" &&
         (prev.type !== "atrule" ||
-          (prev.name !== "import" && prev.name !== "charset"))
+          (prev.name !== "import" && prev.name !== "charset" && prev.name !== "tailwind"))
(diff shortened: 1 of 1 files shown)


RyanZim:
Sorry, we can't be adding support for non-standard at-rules; if we start here, there will be no end to the list. Instead, I suggest you create a file `tailwind-base.css` that simply contains `@tailwind base;`, and likewise for `components`. Then in your main CSS file, you can do: 

```css
@import "./tailwind-base.css";
@import "./tailwind-components.css";

@import "./buttons.css";
```

thunderwp:
OK i understand 
thank you.

PR #456, opened 2021-05-06
remove postcss imports

### LINKED ISSUE #455: Remove postcss imports
According [to migration guide](https://evilmartians.com/chronicles/postcss-8-plugin-migration), the Step 4 recomends to remove the postcss imports and use the second argument of the Root function.

This is not the case of process-content.js file: https://github.com/postcss/postcss-import/blob/master/lib/process-content.js#L7 

I think the `processContent` function should have another argument for that, in order to remove the direct import. This would be very helpful for the Deno version, to avoid loading two different versions of postcss.

oscarotero:
This fixes #455

### DIFF
--- a/index.js
+++ b/index.js
@@ -32,7 +32,7 @@ function AtImport(options) {
 
   return {
     postcssPlugin: "postcss-import",
-    Once(styles, { result, atRule }) {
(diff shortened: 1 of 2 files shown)


RyanZim:
Published v14.0.2

Issue #463, opened 2021-07-11
devDependencies instead of dependencies


EatonZ:
Shouldn't the [installation instructions](https://github.com/postcss/postcss-import/blob/master/README.md#installation) use `npm install -D postcss-import` over `npm install postcss-import` since this is a package used during development, and is not a package containing files that get published?

RyanZim:
Good catch! PR welcome!

EatonZ:
@RyanZim See https://github.com/postcss/postcss-import/pull/464

PR #464, opened 2021-07-12
Update installation instructions to save to devDependencies

EatonZ:
Fixes https://github.com/postcss/postcss-import/issues/463

Since this is a package used during development, and is not a package containing files that get published, it should be a dev dependency.


RyanZim:
Thanks!

PR #465, opened 2021-08-04
[DRAFT] Allows intercepting and skipping the AtRule node

tbranyen:
While postcss-import crawls imports, it is often handy to be able to
mutate the rule before processing. Unfortunately I was unable to
accomplish this by using a plugin before postcss-import. This is due to
rules not being revisited after concatenated with postcss-import.

I also found it odd that there was no way to signal a skip for a given
rule. By signaling a `skip = true` on the node, postcss-imports safely
skips over, but emits a warning. I don't like mutating the AtRule
object with a custom skip property, but this works for now.

Let me know if you need specific background on the motivation for
this change. Opening as draft, since there are no unit tests or
documentation for this change yet. I want to vet this with the team
first to make sure the change is acceptable.

### DIFF
--- a/index.js
+++ b/index.js
@@ -120,7 +120,7 @@ function AtImport(options) {
       }
 
       function parseStyles(result, styles, options, state, media) {
-        const statements = parseStatements(result, styles)
+        const statements = parseStatements(result, styles, options)
 
         return Promise.resolve(statements)
           .then(stmts => {
--- a/lib/parse-statements.js
+++ b/lib/parse-statements.js
@@ -20,14 +20,14 @@ function split(params, start) {
   return list
 }
 
-module.exports = function (result, styles) {
+module.exports = function (result, styles, options) {
   const statements = []
   let nodes = []
 
   styles.each(node => {
     let stmt
     if (node.type === "atrule") {
-      if (node.name === "import") stmt = parseImport(result, node)
+      if (node.name === "import") stmt = parseImport(result, node, options)
       else if (node.name === "media") stmt = parseMedia(result, node)
       else if (node.name === "charset") stmt = parseCharset(result, node)
     }
@@ -78,7 +78,16 @@ function parseCharset(result, atRule) {
   }
 }
 
-function parseImport(result, atRule) {
+function parseImport(result, atRule, options) {
+  if (options.interceptAtRule) {
+    options.interceptAtRule(atRule);
+  }
+  if (atRule.skip) {
+    return result.warn(
+      "@import was skipped by interceptAtRule",
+      { node: atRule }
+    )
+  }
   let prev = atRule.prev()
   if (prev) {
     do {


tbranyen:
@RyanZim I had not tried the `plugins` option within postcss-import, and hoped that might solve the problem. Unfortunately it suffers from the same issue I was facing before. I'm using postcss-loader, less-loader, and file-loader within webpack on an extremely large project with hundreds of entry points. Some of these entry points contain one or more LESS files. Due to legacy reasons these LESS files are not well formed and every `@import` is a `(reference)` import. Like:

```less
@import (reference) './core/variables.less';
```

This will not work with less-loader, since basically nothing is concatenated. My fix is to use postcss to convert these to normal imports by removing the reference modifier.

```js
const replaceImportReferencePlugin = (opts = {}) => {
    return {
        postcssPlugin: 'remove-reference',

        AtRule(rule) {
            // Replace (reference) imports with default.
            if (rule.name === 'import') {
                rule.params = rule.params.replace(/\(reference\) /g, '');
            }
        },
    };
};

replaceImportReferencePlugin.postcss = true;
```

Regardless if the plugin is loaded within postcss-loader or postcss-imports plugins option, I'm facing the following error:

```
Module build failed (from ./node_modules/less-loader/dist/cjs.js):


h1 {
    .style-h1-default();
  ^
.style-h1-default is undefined
```

Which means that the critical variables file was not properly imported. I was able to work around this by using the intercept hook, I've introduced here. My thinking is that while the entry point LESS file is properly processed, this is not working correctly with nested imports within dependencies.

Using this patch, I'm able to get everything correctly imported.

```js
require('postcss-import')({
    interceptAtRule(atRule) {
        // Remove the reference tag.
        atRule.params = atRule.params.replace(
            /\(reference\) /g,
            ''
        );
        // Do not worry about `prev` as `@imports` will be concat in the middle
        // of the file.
        atRule.prev = () => false;
    },
});
```

RyanZim:
Just to be clear, you'd have to use your `replaceImportReferencePlugin` _both_ before `postcss-import`, and passed in the `plugins` option.

Also, we _do_ have an option for skipping imports (I forgot about it): https://github.com/postcss/postcss-import#filter It only gets the filepath though, not the entire import statement.

tbranyen:
Even after adding my plugin in both locations, I get the same error. That was a good thing to try though.

Excerpt from `webpack.config.js`:

```js
postcssOptions: {
    syntax: require('postcss-less'),
    plugins: [
        replaceImportReferencePlugin(),
        require('postcss-import')({
            plugins: [
                replaceImportReferencePlugin(),
            ],
            resolve(uri, base) {
                // This is not a regular module, search in node_modules instead.
                if (!uri.includes('/')) {
                    return require.resolve(uri);
                }
                return path.join(base, uri);
            },
        }),
    ],
},
```

I will try and use the filter option, as that will simplify this PR even more.

tbranyen:
After trying filter:

```js
  filter(filepath) {
      console.log('skip', filepath);
  },
  ```
  
I see an inconsistency mismatch between what shows up in the intercept hook and this function. Like I don't even see the imports I need to filter at all using this filter function. They do show up in the AtRule passed to intercept though.

Before we consider merging this, I will make a reduced test case repo that we can use as a sandbox for this particular problem set.

**Comment by tbranyen** [comment]
Repro here: https://github.com/tbranyen-experimental/postcss-import-repro

**Comment by RyanZim** [comment]
Sorry so long; figured this out. You're using the `AtRule` visitor API, whereas postcss-import uses `Once` with a single walk of the entire css. Turns out, regardless of plugin order, `Once` gets run before the visitor calls. To make your plugin do its transformations first, it will need to use `Once` as well, like this:
```js
const replaceImportReferencePlugin = (opts = {}) => ({
  postcssPlugin: 'remove-reference',

  Once(root) {
    root.walkAtRules((rule) => {
      // Replace (reference) imports with default behavior.
      if (rule.name === 'import') {
        rule.params = rule.params.replace(/\(reference\) /g, '');
      }
    });
  },
});
```
With this change, everything works fine for me. LMK if that works for you.

tbranyen:
Awesome will give that a go! This is the same problem Babel plugins face, where you need to match up the signature to ensure that one runs before the other. I should have tried that first. Thanks for digging in. No worries on the timeline, as I have something working it's not like I was blocked. I'd much prefer using the official API vs introducing something that is probably not needed. :+1:

PR #475, opened 2021-11-18
Support absolute path with root option

Edge00:
Cover the scenario when an `@import` path starts with "/". then the path could be `${options.root}/${import path}`.


RyanZim:
It's not immediately clear to me whether such behavior should be tied to the `root` option, which is primarily about `node_modules` location.

Edge00:
@RyanZim  I get you. However `root` is currently workable with direct (first) absolute import statement (`resolve-abs-root.css` in this commit). Unable to handle indirect absolute import statement (`foo/baz-root.css` imported by `resolve-abs-root.css` in this commit)

Edge00:
@RyanZim  @MoOx  @TrySound  Sorry for disturbing. Could you share some of your opinions abots this?

RyanZim:
> However root is currently workable with direct (first) absolute import statement

It's not in fact working; it happens to work in your test because of https://github.com/postcss/postcss-import/blob/fe000304ed5fe59dd7445b2122f288ea77e5c6b6/lib/resolve-id.js#L30 which makes absolute paths start with `.//`. postcss-import has no actual support for absolute paths.

Edge00:
@RyanZim  Thank you. So you suggest there should be another option field for example `absoluterRoot` to cover my scenario ?

RyanZim:
Yes, it should be a separate option, for sure. The separate question is whether we want to support absolute paths; I feel like this might be a edge case that you should use https://github.com/postcss/postcss-import#resolve for.

Edge00:
OK, Thank you! I'll try https://github.com/postcss/postcss-import#resolve

PR #482, opened 2022-02-11
updated glob link in README

wommy:
old link doesn't work in postcss8, substituted one that does

edit: DOESNT work

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,10 +29,7 @@ a given scope (root, media query...).
 Tests are made from the path & the content of imported files (using a hash
 table).
 If this behavior is not what you want, look at `skipDuplicates` option
-- **If you are looking for glob, or sass like imports (prefixed partials)**,
-please look at
-[postcss-import-ext-glob](https://github.com/dimitrinicolas/postcss-import-ext-glob)
-(which use this plugin under the hood).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -31,7 +31,7 @@ table).
 If this behavior is not what you want, look at `skipDuplicates` option
 - **If you are looking for glob, or sass like imports (prefixed partials)**,
```
RyanZim, line 32: Doesn't look like your plugin supports SASS imports.
```suggestion
- **If you are looking for glob imports**,
```

wommy: pardon my ignorance, as i havent used Sass in many years, 

https://sass-lang.com/guide#topic-4

i feel like the original wording's intention was to catch the attention of sass migrates that used partials,
thats how i stumbled across this

ill have another edit up in a second

wommy: yeah did you read my newer commit?

i reworded the context

RyanZim: Yes; it still mentions migrating from SASS partials; when your plugin does not support anything like SASS partials.

wommy: oh i thought SASS's  @imports had glob functionality,
ill push another

PR #483, opened 2022-02-19
layers support

romainmenke:
_wip_

With `@layer` support soon in all browsers stylesheet authors will also start using these in `@import`.

spec : https://drafts.csswg.org/css-cascade-5/#at-import

new syntax :

```
@import [ <url> | <string> ]
        [ layer |  layer(<layer-name>) ]?
        <import-condition> ;
```

Because I am not familiar with the codebase I first wanted to checkin to make sure I am on the right track.

Status :

- [x] detects and parses `layer`
- [x] failing simple test
- [ ] transform
- [ ] increase test coverage

The intention is to have this transform :

```pcss
@import "foo.css" layer;

/* becomes */

@layer {
  foo {}
}
``` 

I could not easily find the best place to intervene and create a new `atRule` for the layers. Is this something you can help out with @ai?

### DIFF
--- a/lib/parse-statements.js
+++ b/lib/parse-statements.js
@@ -134,11 +134,27 @@ function parseImport(result, atRule) {
   else stmt.uri = params[0].nodes[0].value
   stmt.fullUri = stringify(params[0])
 
-  if (params.length > 2) {
-    if (params[1].type !== "space") {
+  let remainder = params
+  if (remainder.length > 2) {
+    if (
+      (remainder[2].type === "word" || remainder[2].type === "function") &&
+      remainder[2].value === "layer"
+    ) {
+      if (remainder[1].type !== "space") {
+        return result.warn("Invalid import layer statement", { node: atRule })
+      }
+
+      stmt.layer = stringify(remainder[2])
+      remainder = remainder.slice(2)
+    }
+  }
+
+  if (remainder.length > 2) {
+    if (remainder[1].type !== "space") {
       return result.warn("Invalid import media statement", { node: atRule })
     }
-    stmt.media = split(params, 2)
+
+    stmt.media = split(remainder, 2)
   }
 
   return stmt
--- a/test/fixtures/layer.css
+++ b/test/fixtures/layer.css
@@ -1,0 +1,6 @@
+@import "foo.css" layer;
+@import 'bar.css' layer(bar);
+@import url(baz.css) layer;
+@import url("foobar.css") layer(foobar);
+
+content{}
--- a/test/fixtures/layer.expected.css
+++ b/test/fixtures/layer.expected.css
@@ -1,0 +1,14 @@
+@layer {
+  foo {}
+}
+@layer bar {
+  bar {}
+}
+@layer {
+  baz {}
+}
+@layer foobar {
+  foobar {}
+}
+
+content{}
--- a/test/layer.js
+++ b/test/layer.js
@@ -1,0 +1,8 @@
+"use strict"
+// external tooling
+const test = require("ava")
+
+// internal tooling
+const checkFixture = require("./helpers/check-fixture")
+
+test("should resolve layers of import statements", checkFixture, "layer")


RyanZim:
@romainmenke See how we do it for media imports here: https://github.com/postcss/postcss-import/blob/e2d28908a5820a6cf356bb5f8f1ea453169bc693/index.js#L79-L83

romainmenke:
@RyanZim Thank you for the hint!

I think this is ready for review.
Please let me know if anything should be done differently or more test coverage is required.

lib/parse-statements.js
```diff
@@ -85,10 +89,12 @@ function parseImport(result, atRule) {
...
         prev.type !== "comment" &&
         (prev.type !== "atrule" ||
-          (prev.name !== "import" && prev.name !== "charset"))
+          (prev.name !== "import" &&
+            prev.name !== "charset" &&
+            !(prev.name === "layer" && !prev.nodes)))
       ) {
         return result.warn(
-          "@import must precede all other statements (besides @charset)",
+          "@import must precede all other statements (besides @charset or empty @layer)",
```
RyanZim, line 97: Not saying this is wrong, but can you show where the spec allows for empty `@layer` before `@import`?

RyanZim (commented):
One question; otherwise looks good. BTW, sorry so long in getting around to reviewing this.

lib/parse-statements.js
romainmenke: https://drafts.csswg.org/css-cascade-5/#at-import

>Any [@import](https://drafts.csswg.org/css-cascade-5/#at-ruledef-import) rules must precede all other valid at-rules and style rules in a style sheet (ignoring [@charset](https://drafts.csswg.org/css-syntax-3/#at-ruledef-charset) and empty [@layer](https://drafts.csswg.org/css-cascade-5/#at-ruledef-layer) definitions) and must not have any other valid at-rules or style rules between it and previous @import rules, or else the @import rule is invalid.

https://drafts.csswg.org/css-cascade-5/#layer-block

> Note: [@layer](https://drafts.csswg.org/css-cascade-5/#at-ruledef-layer) [block at-rules](https://drafts.csswg.org/css-syntax-3/#block-at-rule) cannot be interleaved with [@import](https://drafts.csswg.org/css-cascade-5/#at-ruledef-import) rules.

In more detail : https://drafts.csswg.org/css-cascade-5/#layer-empty

>The [@layer](https://drafts.csswg.org/css-cascade-5/#at-ruledef-layer) rule can also be used to define new layers without assigning any style rules, by providing only the [layer name](https://drafts.csswg.org/css-cascade-5/#layer-name):
>
>@layer [<layer-name>](https://drafts.csswg.org/css-cascade-5/#typedef-layer-name)[#](https://drafts.csswg.org/css-values-4/#mult-comma);
Such empty [@layer](https://drafts.csswg.org/css-cascade-5/#at-ruledef-layer) rules are allowed before [@import](https://drafts.csswg.org/css-cascade-5/#at-ruledef-import) and @namespace rules (after the [@charset](https://drafts.csswg.org/css-syntax-3/#at-ruledef-charset) rule, if any) as well as everywhere @layer [block at-rules](https://drafts.csswg.org/css-syntax-3/#block-at-rule) are allowed.

------------

But now that I have some distance since writing this I think that `empty @layer` is not a useful way to describe this to stylesheet authors.

_I had to look it up, both to check that I didn't make a mistake and what the exact meaning was._

Is there a better, more widely used name for at-rules without a block?

```css
@layer foo; /* a good name for this? */

@import 'baz.css' layer(baz);
```

RyanZim: I don't know if there's a good name for it; I think following the spec's wording is a safe bet.

romainmenke: That works for me!
We can also see if this causes confusion to authors and adjust if needed.
It might be a non issue as people learn the feature.

RyanZim: All set, just need to merge and release this when I get a chance.

RyanZim:
Just thought about it, perhaps there should be some mention of this in the README?

romainmenke:
Nice! Just used it to double check the example for the readme!

Thank you for all the reviews and guidance 🎉

Issue #495, opened 2022-06-26
@import with layer() places media queries outside of the layer


mvsde:
I created a basic reproduction case here: https://github.com/mvsde/postcss-cascade-layers

What’s happening:

I `@import` another CSS file with `layer(test)`. That other file includes a media query.

After running this through PostCSS with the postcss-import plugin I get the following output:

```css
@layer test {

body {}
}

@media (min-width: 50rem) {
  body {}
}

@layer test {

p {}
}
```

Note how the layer `test` is closed before the media query and “re-opened” afterwards.

I think the expected output should be something like:

```css
@layer test {

body {}

@media (min-width: 50rem) {
  body {}
}

p {}
}
```

romainmenke:
@mvsde Thank you for reporting this!

romainmenke:
@RyanZim I think I can resolve this but not easily without refactoring a substantial part of the code base.

Currently this plugin is written very dense and leverages many javascript mechanics (like `reduce`) that make it much harder to change parts.

Is this something you are open to?

PR #496, opened 2022-06-27
fix import with layer

romainmenke:
fixes : https://github.com/postcss/postcss-import/issues/495

@RyanZim it still feels a bit like the complexity (which I added) is exploding and the overal shape of the code base isn't optimal. But I also do not dislike this fix :)

Important spec bits : https://www.w3.org/TR/css-cascade-5/#at-import

> the optional layer keyword or layer() function assigns the contents of the style sheet into its own anonymous [cascade layer](https://www.w3.org/TR/css-cascade-5/#cascade-layers) or into the named cascade layer.
>
>The layer is added to the [layer order](https://www.w3.org/TR/css-cascade-5/#layer-ordering) even if the import fails to load the stylesheet, but is subject to any [import conditions](https://www.w3.org/TR/css-cascade-5/#import-conditions) (just as if declared by an [@layer](https://www.w3.org/TR/css-cascade-5/#at-ruledef-layer) rule wrapped in the appropriate [conditional group rules](https://www.w3.org/TR/css3-conditional/#conditional-group-rule)).the optional layer keyword or layer() function assigns the contents of the style sheet into its own anonymous [cascade layer](https://www.w3.org/TR/css-cascade-5/#cascade-layers) or into the named cascade layer.
>
>The layer is added to the [layer order](https://www.w3.org/TR/css-cascade-5/#layer-ordering) even if the import fails to load the stylesheet, but is subject to any [import conditions](https://www.w3.org/TR/css-cascade-5/#import-conditions) (just as if declared by an [@layer](https://www.w3.org/TR/css-cascade-5/#at-ruledef-layer) rule wrapped in the appropriate [conditional group rules](https://www.w3.org/TR/css3-conditional/#conditional-group-rule)).

Which means that `@import url('foo.css') layer(A) (min-width: 320px)` becomes :

```css
@media (min-width: 320px) {
  @layer A {
    /* imported styles */
  }
}

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,7 +196,7 @@ Type: `Function`
 Default: `null`
 
 You can provide a custom naming function for anonymous layers (`@import 'baz.css' layer;`).
-This function gets `(index, filename, importRule)` arguments and should return a unique string.
+This function gets `(index, filename)` arguments and should return a unique string.
 
 This option only influences imports without a layer name.
 Without this option the plugin will warn on anonymous layers.
--- a/README.md
+++ b/README.md
@@ -190,6 +190,20 @@ This option is only for adding additional directories to default resolver. If
 you provide your own resolver via the `resolve` configuration option above, then
 this value will be ignored.
 
+#### `nameLayer`
+
+Type: `Function`
+Default: `null`
+
+You can provide a custom naming function for anonymous layers (`@import 'baz.css' layer;`).
+This function gets `(index, filename, importRule)` arguments and should return a unique string.
+
+This option only influences imports without a layer name.
+Without this option the plugin will warn on anonymous layers.
+
+Anonymous layers are very difficult to mimic for this plugin
+but they are equivalent to uniquely named layers.
+
 #### Example with some options
 
 ```js
--- a/README.md
+++ b/README.md
@@ -196,13 +196,10 @@ Type: `Function`
 Default: `null`
 
 You can provide a custom naming function for anonymous layers (`@import 'baz.css' layer;`).
-This function gets `(index, filename)` arguments and should return a unique string.
+This function gets `(index, rootFilename)` arguments and should return a unique string.
 
 This option only influences imports without a layer name.
 Without this option the plugin will warn on anonymous layers.
-
-Anonymous layers are very difficult to mimic for this plugin
-but they are equivalent to uniquely named layers.
 
 #### Example with some options
 
--- a/index.js
+++ b/index.js
@@ -37,6 +37,7 @@ function AtImport(options) {
       const state = {
         importedFiles: {},
         hashFiles: {},
+        anonymousLayerCounter: 0,
       }
 
       if (styles.source && styles.source.input && styles.source.input.file) {
@@ -307,18 +308,28 @@ function AtImport(options) {
       function loadImportContent(result, stmt, filename, options, state) {
         const atRule = stmt.node
         const { media, layer } = stmt
+        if (layer.length === 1 && layer[0] === "") {
+          layer[0] = `importted-anon-layer-${state.anonymousLayerCounter++}`
+        }
+
         if (options.skipDuplicates) {
           // skip files already imported at the same scope
           if (
             state.importedFiles[filename] &&
-            state.importedFiles[filename][media]
+            state.importedFiles[filename][media] &&
+            state.importedFiles[filename][media][layer]
           ) {
             return
           }
 
           // save imported files to skip them next time
-          if (!state.importedFiles[filename]) state.importedFiles[filename] = {}
-          state.importedFiles[filename][media] = true
+          if (!state.importedFiles[filename]) {
+            state.importedFiles[filename] = {}
+          }
+          if (!state.importedFiles[filename][media]) {
+            state.importedFiles[filename][media] = {}
+          }
+          state.importedFiles[filename][media][layer] = true
         }
 
         return Promise.resolve(options.load(filename, options)).then(
@@ -329,8 +340,13 @@ function AtImport(options) {
             }
 
             // skip previous imported files not containing @import rules
-            if (state.hashFiles[content] && state.hashFiles[content][media])
+            if (
+              state.hashFiles[content] &&
+              state.hashFiles[content][media] &&
+              state.hashFiles[content][media][layer]
+            ) {
               return
+            }
 
             return processContent(
               result,
@@ -348,8 +364,13 @@ function AtImport(options) {
                 })
                 if (!hasImport) {
                   // save hash files to skip them next time
-                  if (!state.hashFiles[content]) state.hashFiles[content] = {}
-                  state.hashFiles[content][media] = true
+                  if (!state.hashFiles[content]) {
+                    state.hashFiles[content] = {}
+                  }
+                  if (!state.hashFiles[content][media]) {
+                    state.hashFiles[content][media] = {}
+                  }
+                  state.hashFiles[content][media][layer] = true
                 }
               }
 
--- a/index.js
+++ b/index.js
@@ -308,9 +308,11 @@ function AtImport(options) {
       function loadImportContent(result, stmt, filename, options, state) {
         const atRule = stmt.node
         const { media, layer } = stmt
-        if (layer.length === 1 && layer[0] === "") {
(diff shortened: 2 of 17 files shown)

### CODE REVIEW

**Comment by romainmenke** [comment]
@RyanZim do you have time one of the following days to review this? 🙇

**Comment by RyanZim** [comment]
@romainmenke My apologies for my slowness here; I've been quite busy. Then, I was in a fairly serious accident last week, so that's put things even more on behind. I have not forgotten; I will try to review in the coming days.

File: index.js [hunk 1]
```diff
@@ -79,9 +79,46 @@ function AtImport(options) {
...
+              const layerName = stmt.layer
+                .filter(layer => layer !== "")
+                .join(".")
+
+              let layerParams = "layer"
+              if (layerName) {
+                layerParams = `layer(${layerName})`
+              }
+
+              stmt.node.params = `${stmt.fullUri} ${layerParams})${media}`
```
RyanZim, line 93: Why the closing parenthesis after `layerParams`?

test/fixtures/layer-import-atrules.expected.css
```diff
@@ -0,0 +1,127 @@
...
+
+@layer Y {
+  body {
+    color: purple;
+  }
+}
+}
+}
+
+@media screen {
```
RyanZim, line 22: Why do we close the previous `@media screen` just to reopen a new one right afterwards?

test/fixtures/layer-import-atrules.expected.css
```diff
@@ -0,0 +1,127 @@
...
+@layer imported-with-media {
+
+body {
+  order: 1;
+}
+}
+}
+
+@media screen {
+@layer imported-with-media {
```
RyanZim, line 32: Here again, we close both `@media screen` and `@layer imported-with-media`, only to reopen them.

test/fixtures/layer-import-atrules.expected.css
```diff
@@ -0,0 +1,127 @@
...
+@media (min-width: 50rem) {
+  body {
+    order: 2;
+  }
+}
+}
+}
+
+@media screen {
+@layer imported-with-media {
```
RyanZim, line 43: Again

test/fixtures/layer-import-atrules.expected.css
```diff
@@ -0,0 +1,127 @@
...
+}
+
+@layer imported-without-media {
+
+body {
+  order: 1;
+}
+}
+
+@layer imported-without-media {
```
RyanZim, line 95: Closing and reopening `@layer imported-without-media`

test/fixtures/layer-import-atrules.expected.css
```diff
@@ -0,0 +1,127 @@
...
+
+@media (min-width: 50rem) {
+  body {
+    order: 2;
+  }
+}
+}
+
+@layer imported-without-media {
```
RyanZim, line 104: Again

RyanZim (commented):
I'll be alright, I'm mostly back to normal already.

There's one part of the code I don't understand (left a comment below). Also, it seems there's some unnecessary opening and closing of statements in the output (comments below), is there a reason for this?

index.js
romainmenke: This is the same pattern as we see here : https://github.com/postcss/postcss-import/blob/master/test/fixtures/media-content.expected.css#L3

The `at rules` seem to be closed after each file import.
This is existing behaviour.

test/fixtures/layer-import-atrules.expected.css
romainmenke: This is the same pattern as we see here : https://github.com/postcss/postcss-import/blob/master/test/fixtures/media-content.expected.css#L3

The `at rules` seem to be closed after each file import.
This is existing behaviour.

index.js
romainmenke: This is a typo and after looking into this it is also unreachable code.
I've removed it.

Good catch!

romainmenke:
@RyanZim Thank you for the review and happy to hear things are getting better on your end 🎉

test/fixtures/layer-import-atrules.expected.css
romainmenke: With layers this is indeed an issues if they are anonymous layers.

Opening an closing `@layer {}` multiple times isn't the same opening and closing `@layer foo {}`

We can either de-sugar the anonymous layers and assign them a unique layer name or refactor the plugin further.

romainmenke: I've updated this PR with a fix for anonymous layers.
These are now desugared to `imported-anon-layer-X`

test/fixtures/layer-import-atrules.expected.css
RyanZim: Haha, I'd forgotten how we handled media queries. This is fine then.

RyanZim:
I'm hesitant about the naming of anonymous layers, and what unexpected problems that may create. I'm not super familiar with `@layer` myself, so I don't know all the edge cases.

romainmenke:
> I'm hesitant about the naming of anonymous layers, and what unexpected problems that may create. I'm not super familiar with @layer myself, so I don't know all the edge cases.

I actually am also unsure about this.
Thank you for calling this out.

We could make anonymous layered imports unsupported for now.
That way we avoid unforeseen edge cases while still shipping an improved version.

We can iterate on this later to add support for anonymous layers.
This would only affect `@import "foo.css" layer;` not `@layer {}`.

The current version might be incorrect when an author has multiple stylesheets, all with anonymous layered imports. There is no way for this plugin to be aware of that and the counter approach doesn't produce truly unique layer names.

I did however want to avoid using true random values as these make builds non-reproducible.

RyanZim:
OK, compromise proposal: we don't support anonymous layers by default. To enable support, you must pass an option which is a function, that is passed the index, and must return the layer name. That way, if you want our current index solution, you could pass:

```js
(index) => `imported-anon-layer-${index}`
```

If you have multiple stylesheets, you could use different prefixes. Maybe we should also pass the root filename we're processing, for those who want to incorporate that into the name. Alternately, if someone doesn't care about deterministic builds, they could simply provide a random string generator (with or without a prefix).

I realize this is a bit complicated, but it also provides maximum flexibility; thoughts? Also, sane naming suggestions for such an option?

romainmenke:
I like this!
I will try to implement this in the next few days :)

romainmenke:
@RyanZim I think this is ready for review (when you have time).
For the moment I have settled on `nameLayer` for the plugin option.

Other canidates :

- `nameLayer`
- `deAnonymizeLayer`
- `generateLayerName`

-----

I wasn't sure how to pass the root file name considering that it might not be set.
At the moment I am using the import file name as a second argument.

README.md
```diff
@@ -190,6 +190,20 @@ This option is only for adding additional directories to default resolver. If
...
+Default: `null`
+
+You can provide a custom naming function for anonymous layers (`@import 'baz.css' layer;`).
+This function gets `(index, filename)` arguments and should return a unique string.
+
+This option only influences imports without a layer name.
+Without this option the plugin will warn on anonymous layers.
+
+Anonymous layers are very difficult to mimic for this plugin
+but they are equivalent to uniquely named layers.
```
RyanZim, line 205: This sentence is confusing; I think we're better off removing it.

RyanZim (commented):
> I wasn't sure how to pass the root file name considering that it might not be set.
At the moment I am using the import file name as a second argument.

I think the root file name is way more useful than the import file name. It should be an optional argument; i.e. if the root file name isn't set, `nameLayer` just gets one argument, `index`.

What is the actual output like if you don't pass `layerName`, but use anonymous layers? I'm wondering if we should be erroring instead of warning there: https://github.com/postcss/postcss/blob/main/docs/guidelines/plugin.md#41-use-nodeerror-on-css-relevant-errors

romainmenke:
@RyanZim Thank you for the feedback.
I've made the necessary changes.

Hopefully all good now :)

RyanZim (approved):

RyanZim:
Merged; this is technically semver-major, so I'm going to try to bundle this with a Node version requirement bump to release.

romainmenke:
Thank you for this @RyanZim 🙇

RyanZim:
Released in v15

PR #507, opened 2022-11-11
update test results for PostCSS 8.4.19

romainmenke:
same as https://github.com/postcss/postcss-import/pull/503 but in reverse


RyanZim:
Thanks for following up here!

Issue #509, opened 2022-11-30
Issue with `@layer`s


argyleink:
given
```css
@layer foo {
  :root {
    color: red;
  }
}
@import "icons.css" layer(base);
```

output is missing the contents of icons.css:
```css
@layer foo {
  :root {
    color: red;
  }
}
```

if i remove the layer `foo`, the output is:
```css
@layer base {
.filled-icon {
  color: green;
}
}
```

Seems like there's a bug when there are inline layers already present in the file?

RyanZim:
`@import` statements must come first in CSS files; you should have gotten a warning about this; what postcss runner are you using?

romainmenke:
Confirmed that a warning is emitted by the plugin for this source code.

> `unexpected warning: "@import must precede all other statements (besides @charset or empty @layer)"`

argyleink:
> statements must come first

totally, but, that's what i'm inlining for. once inlined, there's nothing to warn about. plus, why warn and remove the statement? 

> what postcss runner?

npm postcss in deno

<br>
you may be wondering why i have that line of code in my styles? another plugin is importing files from a url into a layer in a transform step before this plugin. i think it'd be best if this plugin didnt care where the import was and instead just replaced it where it found it? or, add the ability to import remote URLs so this plugin can own all the imports and not find itself 2nd in the build step. 

thanks y'all for checking this out tho so quickly, much appreciated!

RyanZim:
> npm postcss in deno

Yeah, you'll want to modify your code to check for `result.warnings` to get the message about problems like this.

This plugin will always follow the CSS spec; when warnings are being omitted because there was invalid input, I don't think it particularly matters what the output is. I'm not sure that a fully understand what you're trying to do, but I see no reason that this plugin needs to change its behavior here.

argyleink:
really, nothing to change? it's removing something entirely and using a warning as a way to not do something? i agree this plugin should follow specs, but not that it should be opinionated in this way. i was assuming the use case i had made it clear of the what and why.. 

i see the other bug was resolved, so maybe i can put this plugin into the transform pipeline in a way that can work around the requirement/warning. but ultimately, i'd prefer this plugin to inline imports, wherever they are. it's fine if it warns about what it did, but my expectation was an import just get inlined. the final output is spec compliant.

romainmenke:
> but my expectation was an import just get inlined. the final output is spec compliant.

I partly agree with @RyanZim here.
[...]

PR #514, opened 2022-12-01
preserve invalid nodes after warnings

romainmenke:
related to : https://github.com/postcss/postcss-import/issues/509

The current implementation removes nodes which are incorrect and emits a warning.
The result is a valid CSS file but its contents are likely not what the author intended.

This change preserves the invalid rules.
The result is now an invalid CSS file but it it can be inspected more easily.
It can be confusing to have a warning about incorrectly ordered `@import` statements if the final output doesn't even contain any `@import`.

I don't have any strong opinion about what is more correct/helpful here.

If error handling is considered part of the API of this plugin then this would be a breaking change.

### DIFF
--- a/lib/parse-statements.js
+++ b/lib/parse-statements.js
@@ -70,9 +70,10 @@ function parseMedia(result, atRule) {
 
 function parseCharset(result, atRule) {
   if (atRule.prev()) {
-    return result.warn("@charset must precede all other statements", {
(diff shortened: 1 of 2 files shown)


romainmenke:
Closing as this doesn't solve anything.
CSS output is invalid with or without this change.

Adding data url support opens more options that this change.

PR #515, opened 2022-12-04
add data url support

romainmenke:
This change adds support for :

```css
@import url(data:text/css;base64,...)
```

_Before this change a data url would throw an error, so this is not something that is being used today by users of this plugin._

As far as I can tell the change required is minimal and everything just works.
- media queries
- at layer
- sourcemaps
- nesting multiple imports within data urls

---------

I don't this is useful on its own.
Authors should not be base64 encoding CSS source code and storing this in `@import` urls.

It can however be useful for other plugins that have non-standard interactions with `@import`.

Another plugin that does want to provide extra features can choose to store as base64 instead of inlining.

This keeps the source valid with all `@import` rules before anything else and all in the correct order.

---------

An alternative approach to this is that other plugins use the file system.
Instead of inlining they can store the contents in a file and update the `@import` rules so they each is fully standard CSS and everything points to a local file.

_Both approaches have merit and I don't have a strong preference between them._

### DIFF
--- a/lib/data-url.js
+++ b/lib/data-url.js
@@ -8,7 +8,7 @@ function isValid(url) {
 
 function contents(url) {
   // "data:text/css;base64,".length == 21
-  return atob(url.slice(21))
+  return Buffer.from(url.slice(21), "base64").toString()
 }
 
 module.exports = {
--- a/lib/data-url.js
+++ b/lib/data-url.js
@@ -7,7 +7,7 @@ function isValid(url) {
 }
 
 function contents(url) {
-  // "data:text/css;base64,".length == 21
+  // "data:text/css;base64,".length === 21
   return Buffer.from(url.slice(21), "base64").toString()
 }
 
--- a/lib/data-url.js
+++ b/lib/data-url.js
@@ -1,0 +1,17 @@
+"use strict"
+
+const dataURLRegexp = /^data:text\/css;base64,/i
+
+function isValid(url) {
+  return dataURLRegexp.test(url)
+}
+
+function contents(url) {
+  // "data:text/css;base64,".length == 21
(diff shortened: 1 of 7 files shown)


lib/data-url.js
```diff
@@ -0,0 +1,17 @@
+"use strict"
+
+const dataURLRegexp = /^data:text\/css;base64,/i
+
+function isValid(url) {
+  return dataURLRegexp.test(url)
+}
+
+function contents(url) {
+  // "data:text/css;base64,".length == 21
```
RyanZim, line 10: I lint comments too :stuck_out_tongue_winking_eye: 
```suggestion
  // "data:text/css;base64,".length === 21
```

RyanZim (approved):
Implementation looks good

RyanZim:
> An alternative approach to this is that other plugins use the file system.
> Instead of inlining they can store the contents in a file and update the @import rules so they each is fully standard CSS and everything points to a local file.

The problem with this approach is that you'd need to do an absolute path to a temporary file, which will create a dependency message for that file, causing runners to watch those files in watch mode, which is undesirable. I think the data URL approach should be preferred for that reason.

Not sure if anyone will use it, but it's not complicated to add, so no real reason not to.

romainmenke:
> The problem with this approach is that you'd need to do an absolute path to a temporary file, which will create a dependency message for that file, causing runners to watch those files in watch mode, which is undesirable. I think the data URL approach should be preferred for that reason.

Cool! Thank you reviewing this 🙇

RyanZim:
Published in `postcss-import@15.1.0`

romainmenke:
Thank you for the release @RyanZim  🙇

Issue #519, opened 2023-02-08
Document that all forms of `url` are supported


romainmenke:
https://developer.mozilla.org/en-US/docs/Web/CSS/@import

> Is a [<string>](https://developer.mozilla.org/en-US/docs/Web/CSS/string), a <url>, or a [url()](https://developer.mozilla.org/en-US/docs/Web/CSS/url) function representing the location of the resource to import. The URL may be absolute or relative.

PR #520, opened 2023-02-08
add docs for url notation

romainmenke:
fixes : https://github.com/postcss/postcss-import/issues/519

We got a question about this in the CSSTools discord.
It seems that people assume that `@import url()` is not supported because it isn't listed in the readme.

This change adds some simple examples.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,17 +74,20 @@ postcss()
 `css/input.css`:
 
 ```css
+/* remote urls are preserved */
+@import "https://example.com/styles.css";
+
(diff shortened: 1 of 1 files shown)


RyanZim (changes requested):
Need to edit the example output as well to add some comment about "content of foo.css"

Also, perhaps we should explicitly show that remote URLs are passed through?

romainmenke:
Thank you for the feedback @RyanZim.
I've made a few more edits.

RyanZim (approved):

PR #530, opened 2023-07-14
add parsing support for supports conditions

romainmenke:
- added parsing support for `supports` conditions
- rewrote the import statement parser
- added coverage checks with `c8`

---------

Fully implementing supports conditions will require a fairly large refactor and the codebase is already very complex.

I would like to first do a few refactors.
This is one of them.

Initially the parsing algorithm was a simple split.
Look for anything that could be a uri, everything after is a `media` condition.

With `layer` and `supports` something more flexible is needed.

Any `supports` conditions will currently give errors.

This is already a behavior change.
Before this PR `supports` conditions would be treated as if they were `media` conditions, which is always incorrect.

Both behaviors are broken but after this change CSS authors will receive more meaningful feedback.

### DIFF
--- a/lib/parse-statements.js
+++ b/lib/parse-statements.js
@@ -113,59 +113,97 @@ function parseImport(result, atRule) {
   const params = valueParser(atRule.params).nodes
   const stmt = {
     type: "import",
+    uri: "",
+    fullUri: "",
     node: atRule,
     media: [],
     layer: [],
+    supports: [],
   }
 
-  // prettier-ignore
-  if (
-    !params.length ||
-    (
-      params[0].type !== "string" ||
-      !params[0].value
-    ) &&
-    (
-      params[0].type !== "function" ||
-      params[0].value !== "url" ||
-      !params[0].nodes.length ||
-      !params[0].nodes[0].value
-    )
-  ) {
-    return result.warn(`Unable to find uri in '${  atRule.toString()  }'`, {
+  for (let i = 0; i < params.length; i++) {
+    const node = params[i]
+
+    if (node.type === "space" || node.type === "comment") continue
+
+    if (node.type === "string") {
+      if (stmt.uri) {
+        return result.warn(`Multiple url's in '${atRule.toString()}'`, {
+          node: atRule,
+        })
+      }
+
+      if (!node.value) {
+        return result.warn(`Unable to find uri in '${atRule.toString()}'`, {
+          node: atRule,
+        })
+      }
+
+      stmt.uri = node.value
+      stmt.fullUri = stringify(node)
+      continue
+    }
+
+    if (node.type === "function" && /^url$/i.test(node.value)) {
+      if (stmt.uri) {
+        return result.warn(`Multiple url's in '${atRule.toString()}'`, {
+          node: atRule,
+        })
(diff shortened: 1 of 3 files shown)


RyanZim:
@romainmenke are you thinking we should release with this warning for `@supports`, or is this just a temporary measure until we add support after refactoring?

lib/parse-statements.js
```diff
@@ -113,59 +113,97 @@ function parseImport(result, atRule) {
...
-      !params[0].nodes.length ||
-      !params[0].nodes[0].value
-    )
-  ) {
-    return result.warn(`Unable to find uri in '${  atRule.toString()  }'`, {
-      node: atRule,
-    })
-  }
+  for (let i = 0; i < params.length; i++) {
+    const node = params[i]
```
RyanZim, line 125: Maybe should use a for..of loop here?

lib/parse-statements.js
```diff
@@ -113,59 +113,97 @@ function parseImport(result, atRule) {
...
+      stmt.uri = node.nodes[0].value
+      stmt.fullUri = stringify(node)
+      continue
+    }
+
+    if (!stmt.uri) {
+      return result.warn(`Unable to find uri in '${atRule.toString()}'`, {
+        node: atRule,
+      })
+    }
```
RyanZim, line 169: This code is placed here, and again further below, why?

lib/parse-statements.js
romainmenke: We need the index on line 188 for `split(params, i)`

lib/parse-statements.js
romainmenke: The first one, inside the `for`, is an early abort in case something that can not be a url is found before a url. `@import screen url("foo");`

The second is to make sure that any url was found. `@import screen;`

The first one could be dropped, it's not needed functionally, it's only an early return.

romainmenke:
> are you thinking we should release with this warning for @supports, or is this just a temporary measure until we add support after refactoring?

I am hoping it is temporary.
But I want to keep the code base healthy after each change and I would like to keep each change small-isch.

The state in this pull request is something that can be released if needed.

I plan to continue work towards full support for `supports` conditions, but this could take a while. It really complicates things that four different types of rules interact (`import`, `media`, `layer`, `supports`). With `layer` I was able to patch a few existing bits and make it work, but a larger refactor is needed for `supports`.

lib/parse-statements.js
RyanZim: Ah, missed that the nesting level was different. Goes to show I need to review this in more detail when less tired...

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# Unreleased
+
+- Give a warning when `supports` conditions are used in `@import` statements.
+
```
RyanZim, line 4: It's not yet clear if we'll be releasing this change; I don't mind writing up the changelogs at release time. Let's just remove this.

RyanZim (commented):
Let's split the addition of c8 into its own PR.

Code here LGTM.

romainmenke:
Thank you for the review 🙇

PR #534, opened 2023-07-20
extend and fix data url imports

romainmenke:
- add support for url encoded data urls
- fix handling of embedded relative imports

### DIFF
--- a/lib/parse-styles.js
+++ b/lib/parse-styles.js
@@ -17,9 +17,10 @@ async function parseStyles(
   state,
   media,
   layer,
+  from,
   postcss
 ) {
-  const statements = parseStatements(result, styles)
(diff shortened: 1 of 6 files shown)


lib/parse-styles.js
```diff
@@ -93,6 +94,10 @@ async function resolveImportId(result, stmt, options, state, postcss) {
     return
   }
 
+  if (dataURL.isValid(stmt.from)) {
+    return
+  }
```
RyanZim, line 99: Perhaps leave a comment on what we're doing here, and why? Won't be obvious to our future selves.

RyanZim (approved):

romainmenke:
Thank you for the review 🙇

Issue #462, opened 2021-06-23
Duplicate imports are resolved in the wrong order


evanw:
This plugin uses a different import order than the browser, which can lead to visual discrepancies when CSS is bundled vs. unbundled. I believe this is a regression and that it was caused by #211. The behavior used to be correct but was broken in that change, and is now incorrect.

Here's an example (note that the file `common.css` is being imported twice):

```
$ cat common.css
body { background: white; color: black; }

$ cat a.css
@import "./common.css";
body { background: red; }

$ cat b.css
@import "./common.css";
body { color: red; }

$ cat entry.css
@import "./a.css";
@import "./b.css";

$ cat index.html
<link rel="stylesheet" href="entry.css">
Some example text
```

When you view `index.html` in a browser, you should see a white background and red text. However, when bundled with PostCSS, the entire page is red.

Here's the script I used to bundle with PostCSS:

```js
const css = require('fs').readFileSync('entry.css', 'utf8')
require('postcss')().use(require('postcss-import')({
})).process(css, {
  from: 'entry.css',
}).then(result => {
  console.log(result.css)
})
```

The incorrect behavior can be "fixed" by adding `skipDuplicates: false` to the `postcss-import` options, but that's not great because it makes the output bigger unnecessarily. This option is also dangerous because `skipDuplicates: false` can cause `postcss-import` to infinite-loop and hang if the import graph has cycles.

The way to fix all of these problems and have `postcss-import` match what the browser does is to only respect the _last_ duplicate `@import` statement instead of only respecting the _first_ duplicate `@import` statement.

RyanZim:
Sorry for the slow reply here; reproduced. However, `postcss-import` has had this incorrect behavior for a long time; changing it completely would cause a massive shockwave in the ecosystem. Also, depending what plugins you're using, for example https://github.com/postcss/postcss-simple-vars, you _want_ the current behavior, so your imported files are always at the top of the file.

At this point, I think our only option is to introduce an option to specify whether you want to use the first or last duplicate import. Perhaps roll `skipDuplicates` into this as a new single option for specifying duplicate handling. Default would be debatable.

marat-gainullin:
[...]

PR #535, opened 2023-07-20
check import cycles

romainmenke:
resolves : https://github.com/postcss/postcss-import/issues/462

It is untrue that browsers only use the last import.
They use all of them.

Up until now however this wasn't observable.

With the introduction of `@layer` CSS now has something that is registered when first encountered and can not be overwritten later, making the real behavior clear.

When `skipDuplicates: false` is set `postcss-import` will produce CSS that is equivalent to browser behavior.

This change adds handling of cyclical imports in the same way as browsers resolve these cycles. This removes any differences between bundled or unbundled CSS as affected by import order.

-----

See WebKit implementation : https://github.com/WebKit/WebKit/blob/main/Source/WebCore/css/StyleRuleImport.cpp#L123-L131


lib/parse-styles.js
```diff
@@ -165,6 +165,8 @@ async function loadImportContent(
       state.importedFiles[filename][media] = {}
     }
     state.importedFiles[filename][media][layer] = true
+  } else if (from.includes(filename)) {
+    return
   }
```
romainmenke, line 170:
```suggestion
  }

  if (from.includes(filename)) {
    return
  }
```

Even with `skipDuplicates: true` there can be import cycles when they have alternating conditions.

```css
/* a.css */
@import url("b.css") screen;
```

```css
/* b.css */
@import url("a.css") all;
```

The conditions will always be joined with `and` creating an ever longer but always unique condition : `screen and all and screen and all and ...`

romainmenke: needs a test case

RyanZim (approved):
Sorry so slow in reviewing here; looks good.

PR #538, opened 2023-07-22
fix media query joins

romainmenke:
When the import condition is the same the condition was erased.


RyanZim (approved):

RyanZim:
Sorry for my slowness in reviewing of late; life has a way of getting busy. :grinning:

romainmenke:
> Sorry for my slowness in reviewing of late

That is totally fine :D
I don't mind if things like this hang for a few weeks.

Thank you for reviewing 🙇

Issue #537, opened 2023-07-21
URL's with a fragment or query aren't skipped and give a file not found error


romainmenke:
```css
@import "foo.css#bar";
@import "foo.css?bar=baz";
```

These should be skipped for the same reason that url's with a protocol are skipped.

-----

somewhat related to https://github.com/postcss/postcss-import/issues/536

romainmenke:
Can be done with :

```js
function isProcessableURL(uri) {
  if (/^(?:[a-z]+:)?\/\//i.test(uri)) {
    return false
  }

  try {
    // needs a base to parse properly
    const url = new URL(uri, "https://example.com")

    if (url.hash) {
      return false
    }

    if (url.search) {
      return false
    }
  } catch {} // Ignore

  return true
}
```

PR #539, opened 2023-07-23
ignore/skip urls with a fragment or query

romainmenke:
fixes : https://github.com/postcss/postcss-import/issues/537

### DIFF
--- a/lib/parse-styles.js
+++ b/lib/parse-styles.js
@@ -27,7 +27,6 @@ async function parseStyles(
     stmt.parentMedia = media
     stmt.layer = joinLayer(layer, stmt.layer || [])
 
-    // skip protocol base uri (protocol://url) or protocol-relative
     if (stmt.type !== "import" || !isProcessableURL(stmt.uri)) {
       continue
     }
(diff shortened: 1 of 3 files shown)


lib/parse-styles.js
```diff
@@ -28,7 +28,7 @@ async function parseStyles(
     stmt.layer = joinLayer(layer, stmt.layer || [])
 
     // skip protocol base uri (protocol://url) or protocol-relative
```
RyanZim, line 30:
```suggestion
```

lib/parse-styles.js
```diff
@@ -219,4 +219,25 @@ async function loadImportContent(
   )
 }
 
+function isProcessableURL(uri) {
+  if (/^(?:[a-z]+:)?\/\//i.test(uri)) {
```
RyanZim, line 223:
```suggestion
  // skip protocol base uri (protocol://url) or protocol-relative
  if (/^(?:[a-z]+:)?\/\//i.test(uri)) {
```

lib/parse-styles.js
```diff
@@ -219,4 +219,25 @@ async function loadImportContent(
...
+    return false
+  }
+
+  try {
```
RyanZim, line 227:
```suggestion
  // check for fragment or query 
  try {
```

RyanZim (approved):
Suggested a few comment changes

romainmenke:
Thank you for the feedback and suggestions!

PR #543, opened 2023-08-09
Error on embedded imports in data urls

romainmenke:
The current code prevents further inlining of embedded imports but it doesn't stop a browser from possible loading and applying it.

~~Adding `not all` has the intended effect.~~

Setting `stmt.children = []` prevents inlining and removes the `@import` statement.

### DIFF
--- a/test/fixtures/data-url.expected.css
+++ b/test/fixtures/data-url.expected.css
@@ -15,10 +15,6 @@ p { color: blue; } } }
 
 /* Mixed imports: */
 
-p {
-  color: blue;
-}
-
 p { color: pink; }
 
 /* url encoded: */
--- a/test/fixtures/data-url.expected.css
+++ b/test/fixtures/data-url.expected.css
@@ -1,4 +1,4 @@
-@import url(foo.css);p { color: green; }p { color: blue; }@media (min-width: 320px){@layer foo{
+@import url(foo.css) not all;p { color: green; }p { color: blue; }@media (min-width: 320px){@layer foo{
 p { color: green; } } }@media (min-width: 320px){@layer foo{
 
 p { color: blue; } } }/* Mixed imports: */p {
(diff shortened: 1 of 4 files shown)


RyanZim:
If such imports are completely useless, why do we output useless imports vs. erroring?

romainmenke:
A warning would help users to write more effective CSS.

But a hard error might be too much as this is perfectly valid CSS, it's just CSS that isn't applied.

This is similar to the warning for empty files.
Empty files are valid, just not very useful.

romainmenke:
I've added a suggestion for a warning, please let me know if this is clear.

-----

An alternative would be to warn and completely remove the `@import` node.

But that would require a larger refactor, adding `not all` is a non-intrusive and technically correct way to ignore these `@import` statements.

test/fixtures/data-url.expected.css
```diff
@@ -1,6 +1,28 @@
-@import url(foo.css);p { color: green; }p { color: blue; }@media (min-width: 320px){@layer foo{
-p { color: green; } } }@media (min-width: 320px){@layer foo{
+p { color: green; }
```
romainmenke, line 1: This change affects the output format because the first node is removed.

RyanZim:
Needs rebase.

> But a hard error might be too much as this is perfectly valid CSS, it's just CSS that isn't applied.

In other words, you're saying browsers ignore imports like these? I'm not sure if that's a valid premise to build on, because browsers will continue to parse the CSS they can, even if an import results in a 404; however, as a build tool, we'd be remiss not to error out if we can't resolve an import.

romainmenke:
> we'd be remiss not to error out if we can't resolve an import.

Makes sense!
I've changed it to an error.

Thank you for reviewing these

RyanZim (approved):

Issue #529, opened 2023-07-14
Supports at rule not working


lubomirblazekcz:
```css
@import "style.css" supports(not (color: color-mix(in srgb, red 0%, red)));
```

Transforms into 

```css
@media supports (not (color: color-mix(in srgb,red 0%,red)))
```

And it should be

```css
@supports (not (color: color-mix(in srgb,red 0%,red)))
```

PR #548, opened 2023-12-09
implement `supports` conditions

romainmenke:
supersedes : https://github.com/postcss/postcss-import/pull/532
fixes : https://github.com/postcss/postcss-import/issues/529

----------

The core issue is how multiple statements are combined.

The previous implementation tried to squash media queries and layer conditions.

- `layer(foo) screen`
- `layer(bar) (min-width: 1000px)`

becomes :
```css
@media screen and (min-width: 1000px) {
  @layer foo.bar {}
}
```

In this example the `foo` layer is only defined when the media is `screen` and has a width of `1000px`. While it should actually be defined even when the media is `screen` and smaller.

- `screen`
- `not print`

becomes :
```css
@media screen and not print {}
```

In this example the media query becomes invalid because `and` can not be used to combine `screen` and `not print`

-------

The implementation for this feature was also fairly complex.

Adding `supports` into this would have made the implementation very unwieldy and it would have increased the chance that users experience bugs.

-------

The correct way to combine is actually much simpler.

- `layer(foo) screen`
- `layer(bar) (min-width: 1000px)`

becomes :
```css
@media screen {
  @layer foo {
    @media (min-width: 1000px) {
      @layer foo.bar {}
    }
  }
}
```

- `screen`
- `not print`

becomes :
```css
@media screen {
  @media not print {}
}
```

--------

By using nested statements we ensure that the order of declaration is respected and that no invalid statements are produced.

At this point it also becomes trivial to add support for `supports` conditions.

--------

This method does break one aspect.

`@import` statements can not be nested syntactically in CSS.
They must appear first at the root of the AST.

valid :
```css
@imports "http://something-external.com" screen;
````

invalid :
```css
@media screen {
  @imports "http://something-external.com";
}
````

To restore support for this we can use nested stylesheets.
`@import` statement urls can be base64 encoded data urls.
This doesn't actually have to be base64 encoded, but encoding makes it less prone to errors.

Without encoding it looks like this :

```css
@import url("data:text/css,@imports url('http://something-external.com') not print;") screen;
```

- import this : `@imports url('http://something-external.com') not print;`
- apply a `screen` media condition

_This technique was first thought of by the maintainer of esbuild while fixing issues brought to light by https://github.com/romainmenke/css-import-tests_

--------------

This implementation change fixes a lot of subtle bugs and as far as I can tell the output is almost always smaller.

It also eliminates some complexity for users.
They no longer need to define things like `nameLayer` for anonymous layers to work correctly. The issue for which we introduced that option no longer exists in this implementation.

So I think this is a pure win :)


test/layer.js
```diff
@@ -17,66 +10,13 @@ test("should resolve layers of import statements", checkFixture, "layer")
...
-  // A stable, deterministic and unique layer name:
-  // - layer index
-  // - relative rootFilename to current working directory
-  return `import-anon-layer-${crypto
-    .createHash("sha256")
-    .update(`${index}-${rootFilename.split(cwd)[1]}`)
-    .digest("hex")
-    .slice(0, 12)}`
-}
+test("should group rules", checkFixture, "layer-rule-grouping")
```
RyanZim, line 22: Is this description still accurate?

Also https://github.com/postcss/postcss-import/blob/3d51fe5009dbaccf7e4d1bdac75a1d7478c36f00/test/media.js#L28

lib/process-content.js
```diff
@@ -21,7 +21,6 @@ module.exports = function processContent(
   // SugarSS support:
   if (ext === ".sss") {
     if (!sugarss) {
-      /* c8 ignore next 3 */
```
RyanZim, line 24: Why this change?

RyanZim (commented):
Just to clarify, base64 encoding is only used for external URL imports, right?

A few questions below as well; otherwise, looks good.

lib/process-content.js
romainmenke: Unsure what my motivation was at the time.
I've reverted this to the previous state.

test/layer.js
romainmenke: ### layer

That entire test isn't that relevant anymore because the mechanic it was testing no longer exists.

If I remove it the test coverage also remains the same.

It does however contain a seemingly unrelated part.
This test has duplicate anonymous layered imports which merit a dedicated test.

I've added a test with and without `skipDuplicates` for duplicate anonymous layered imports and I've removed the `layer-rule-grouping` test.

--------

### media

I've relabeled this test from `media-join` to `media-combine`.

I think `join` is too easily associated with an array join into a string.

`combine` is more general and remains correct no matter the underlying method (nesting at rules vs. concatenating at rule params)

romainmenke:
Thank you for the review 🙇 

> Just to clarify, base64 encoding is only used for external URL imports, right?

Yes. 
Only when an import can not be inlined and conditions or cascade layers need to be applied.

RyanZim (approved):

RyanZim:
Thanks a lot! Ready to publish as v16?

RyanZim:
I think I'll wait until after the New Year to publish, just in case something goes wrong.

RyanZim:
https://github.com/postcss/postcss-import/commit/025f27a86018d052d1de62bfe636931bc0f28a2f :tada:

romainmenke:
Awesome! thank you so much for all the careful reviews 🙇

Issue #557, opened 2024-02-09
Emitted `@media` AtRules are invalid


sandhose:
`@import ... media queries;` statements emit invalid `AtRule`, as the `AtRule.params` is an array instead of being a string.

This breaks the tailwindcss plugin in some scenarios, like:

```
@import url('./dark.css') screen and (prefered-color-scheme: dark);
@import url('./light.css') screen and (prefered-color-scheme: light);
```

Compiling this with both the postcss-import and tailwindcss plugin results to:

```
TypeError: (intermediate value)(intermediate value)(intermediate value).replace is not a function
    at ./node_modules/tailwindcss/lib/lib/collapseAdjacentRules.js:37:149
    at Array.every (<anonymous>)
    at ./node_modules/tailwindcss/lib/lib/collapseAdjacentRules.js:37:35
    at Root.each (./node_modules/postcss/lib/container.js:53:16)
    at collapseRulesIn (./node_modules/tailwindcss/lib/lib/collapseAdjacentRules.js:24:14)
    at ./node_modules/tailwindcss/lib/lib/collapseAdjacentRules.js:59:9
    at ./node_modules/tailwindcss/lib/processTailwindFeatures.js:61:53
    at async plugins (./node_modules/tailwindcss/lib/plugin.js:38:17)
    at async LazyResult.runAsync (./node_modules/postcss/lib/lazy-result.js:261:11)
    at async Promise.all (index 0)
```

I've done a minimal example to reproduce this issue here: https://github.com/sandhose/postcss-import-bug-repro

This bug was introduced in version 16.0.0

romainmenke:
@sandhose Thank you for reporting this.
I will take a look shortly!

romainmenke:
I've prepared a fix here : https://github.com/postcss/postcss-import/pull/558
Thank you again for the reproduction.

sandhose:
You're welcome! Thanks for your really quick response!

RyanZim:
Published in v16.0.1 :fireworks:

PR #558, opened 2024-02-09
ensure that media params remain string values

romainmenke:
fixes : https://github.com/postcss/postcss-import/issues/557

### DIFF
--- a/lib/parse-statements.js
+++ b/lib/parse-statements.js
@@ -6,20 +6,6 @@ const valueParser = require("postcss-value-parser")
 // extended tooling
 const { stringify } = valueParser
 
-function split(params, start) {
-  const list = []
-  const last = params.reduce((item, node, index) => {
-    if (index < start) return ""
-    if (node.type === "div" && node.value === ",") {
-      list.push(item)
-      return ""
-    }
-    return item + stringify(node)
-  }, "")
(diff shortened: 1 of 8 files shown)


lib/parse-statements.js
```diff
@@ -6,20 +6,6 @@ const valueParser = require("postcss-value-parser")
...
-    if (index < start) return ""
-    if (node.type === "div" && node.value === ",") {
-      list.push(item)
-      return ""
-    }
-    return item + stringify(node)
-  }, "")
-  list.push(last)
-  return list
-}
```
romainmenke, line 21: I overlooked that this was still present and JavaScript automatically joins arrays of strings with `,`.

This entire section has become redundant with the refactors for `v16`.

RyanZim (approved):
Code looks good; should we add a test case with the reproduction from https://github.com/postcss/postcss-import/issues/557?

romainmenke:
Thank you for reviewing @RyanZim 🙇 

I've added a separate plugin that always runs after `postcss-import` and checks the types of all `AtRule` AST nodes. This way we get maximum coverage for this aspect as it runs with each test case.

It tests a few more things than just `atrule.params`, just in case.

RyanZim (approved):

PR #561, opened 2024-02-20
Allow urls that contain fragments, fixes #560

### LINKED ISSUE #560: Support a way to process a URL with fragments
In Vite, we were resolving `@import '#foo'` with [subpath imports](https://nodejs.org/api/packages.html#subpath-imports). But since postcss 16.0.0 (https://github.com/postcss/postcss-import/pull/539), these URLs are filtered out before it reaches any code that the plugin user side can control. That made Vite not able to resolve these URLs (https://github.com/vitejs/vite/issues/15861).

Would it be possible to add an option to skip this default behavior?

romainmenke:
The intention behind ignoring urls with a fragment was to make the plugin more correct.
The thinking at the time was that these are never valid imports for a filesystem.
With nodejs subpath imports that assumption is incorrect.

### DIFF
--- a/lib/parse-styles.js
+++ b/lib/parse-styles.js
@@ -227,11 +227,6 @@ function isProcessableURL(uri) {
   try {
     // needs a base to parse properly
     const url = new URL(uri, "https://example.com")
-
-    if (url.hash) {
-      return false
-    }
-
     if (url.search) {
       return false
     }
--- a/test/fixtures/filter-ignore.css
+++ b/test/fixtures/filter-ignore.css
@@ -13,5 +13,4 @@
 @import url('//css');
 @import url(//css);
 @import url('foo.css?query=string');
-@import url('foo.css#hash');
 content{}
--- a/test/fixtures/filter-ignore.expected.css
+++ b/test/fixtures/filter-ignore.expected.css
@@ -14,7 +14,6 @@
 @import url('//css');
 @import url(//css);
 @import url('foo.css?query=string');
-@import url('foo.css#hash');
 @media (min-width: 25em){
 ignore{}
 }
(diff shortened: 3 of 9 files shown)


sapphi-red (approved):
Thank you! This test case covers the Vite use case 👍

romainmenke:
@RyanZim could you take a look at this when you have time?
It is not very urgent given that Vite applies a patch for now, but I don't want to forget about this :) Thank you

RyanZim:
Apologies, I've been incredibly busy. This is technically a breaking change, right? Code LGTM.

romainmenke:
I don't think this is a breaking change :)

Import statements with fragments were ignored before this change.
Effectively resulting in left over `@import` in the middle of your CSS file.
This would be ignored by browsers as `@import` needs to come first.

This would be a silent failure.
But it is extremely unlikely that users started doing this since `v14` was released.

After this change `postcss-import` will try to resolve the file.
Unless you have special handling, like Vite does, you will get a hard error on a missing file.

This is no longer a silent failure but users can trivially resolve this by removing the incorrect `@import`.

I changed it in v14 was because it is more correct in a bundler context when only considering the CSS specification. I wasn't aware of how tools like Vite used this.

RyanZim:
Thanks for the detailed explanation. I think it's still technically a breaking change, key word being _technically_, but I agree it probably doesn't affect any real-world use, so probably safe to release without a major bump.

RyanZim:
I'm a fairly strong believer in "never ship on Fridays," so will be sometime next week until this is released.

RyanZim:
Published v16.1.0

romainmenke:
Thank you @RyanZim 🙇

Issue #567, opened 2024-10-18
`@import` with `layer` gets reordered before `@layer` and changes the behavior


sapphi-red:
```css
@layer reset, bootstrap;
@import url('https://cdn.jsdelivr.net/npm/bootstrap@5.3.0/dist/css/bootstrap.css') layer(bootstrap);

@layer reset {
  h1 {
    margin-top: 100px;
  }
}
```
This input is transformed into:
```css
@import url('https://cdn.jsdelivr.net/npm/bootstrap@5.3.0/dist/css/bootstrap.css') layer(bootstrap);
@layer reset, bootstrap;
@layer reset {
  h1 {
    margin-top: 100px;
  }
}
```
But the output CSS has a different meaning from the input CSS (`bootstrap` layer is now overridden with `reset` layer).

Reproduction: https://stackblitz.com/edit/template-sapphi-red-postcss-sass-kasrpt?file=src%2Finput.css,postcss.js

romainmenke:
Thank you for reporting this @sapphi-red 

This is a rather complex issue and will require a substantial rewrite.
Is this an effort that Vite could support financially?

----

fyi: this issue doesn't exist in [`@csstools/postcss-bundler`](https://github.com/csstools/postcss-plugins/tree/main/plugin-packs/postcss-bundler#postcss-bundler-)

But there I didn't implement a lot of non-standard behavior nor did I provide any of the customization options that exist in `postcss-import`.

`@csstools/postcss-bundler` isn't suitable for Vite but it does show that this issue is solvable :)

sapphi-red:
Hi, we had sponsored [the PostCSS organization through VoidZero via OpenCollective](https://opencollective.com/postcss) last month to collectively support the PostCSS ecosystem, which I think includes this repo too. We avoid providing financial support for specific issues or pull requests. We do this to avoid the possibility that financial support might cause merging to become a goal in itself, which would remove the long-term perspective and put pressure on library maintainers. Therefore, we prefer to provide financial support to library organizations. I hope you're able to claim your refactoring efforts from OpenCollective. Thanks for your work here!

stevegreco:
It seems like this makes layers almost completely useless when using `postcss-import`.  I have adopters of our internal Design System trying to integrate our CSS into their apps and there is no way to do it.

This seems to end up breaking everything.

```css
@layer base, components, designsystem, utilities;

@import "tailwindcss/base" layer(base);
@import "tailwindcss/components" layer(components);
[...]

PR #574, opened 2025-06-02
Fix incorrect cascade layer order when some resources can not be inlined

romainmenke:
Fixes the original problem described in https://github.com/postcss/postcss-import/issues/567#issue-2596115833

The root cause is that all nodes are added to either the `imports` or `bundle` array in https://github.com/postcss/postcss-import/blob/d43ca1506d65d9943a3c4bb885aff94dc809d69c/lib/parse-styles.js#L36-L37

While `@layer` statements (those without a rule body) can precede `@import`.
To fix this issue I've added a new statement type for layers and added the needed mechanics to apply conditions to these. To express conditional `@layer` statements before other `@import` statements, or `@layer` statements that end up in between other `@import` statements we use base64 encoded stylesheets:

`@import 'data:text/css;base64,QGxheWVyIGxheWVyLWJldGE=' print;` is equivalent to `@media print { @layer layer-beta }` but is valid before other `@import` statements

I am unsure if it fixes all other reported examples as I don't have reproductions for those. I suggest they test out the fix (if/when it lands) and file new issues if they still encounter problems.

I verified the fix by re-testing with https://github.com/romainmenke/css-import-tests
With the patch applied 4 more test cases pass, without regressions in other cases.

### DIFF
--- a/lib/apply-conditions.js
+++ b/lib/apply-conditions.js
@@ -3,12 +3,38 @@
 const base64EncodedConditionalImport = require("./base64-encoded-import")
 
 module.exports = function applyConditions(bundle, atRule) {
-  bundle.forEach(stmt => {
+  const firstImportStatementIndex = bundle.findIndex(
+    stmt => stmt.type === "import",
+  )
+  const lastImportStatementIndex = bundle.findLastIndex(
+    stmt => stmt.type === "import",
+  )
+
+  bundle.forEach((stmt, index) => {
+    if (stmt.type === "charset" || stmt.type === "warning") {
+      return
+    }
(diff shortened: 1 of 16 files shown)


RyanZim:
Will release next week

RyanZim:
Actually, there's no good reason for me to be the bottleneck for npm releases; @romainmenke I should add you on npm; just to confirm, you're https://www.npmjs.com/~romainmenke there?

RyanZim:
@romainmenke Invitation sent; feel free to push the version bump/CHANGELOG update directly to master.

PR #598, opened 2026-06-30
Clarify that `@layer` can be used before `@import`

o-t-w:
Update README to clarify that `@layer` can be used before `@import`


romainmenke:
@o-t-w Can you update your commit message so that at rule names are either quoted or written as `at-layer`, `layer` or ...

Otherwise we ping unrelated persons or orgs and associated them with commits 😅

RyanZim:
@romainmenke we can edit the commit message when merging

